PR #1, opened 2020-11-05
Index rewrite

arj03:
<blank>


staltz:
Idea: aim to have all secret-stack plugins under a namespace. If the global namespace is unavoidable, put those methods in a dedicated plugin, e.g. `require('ssb-db2/compat')` which defines global namespace methods. Try to have `compat` be a thin wrapper for methods implemented elsewhere.

staltz:
Question: in what module should we call `db2.registerPlugin()`?

arj03:
I would imagine that db2 by default only includes base index. And then you just add a `require('ssb-db2/social')` that would include and register the social indexes. This also allows it to be moved to its own module.

arj03:
node 14 failure seems to be related to ssb-keys 8 using an old version of chloride. See [this](https://github.com/ssb-js/ssb-keys/pull/84#issuecomment-699845253)

arj03:
The node 14 thing just means that it is slow to run the CI. Not a big problem.

arj03:
Reviewed with @staltz

PR #10, opened 2020-11-13
move operators from jitdb to ssb-db2

staltz:
To be merged simultaneously with https://github.com/ssb-ngi-pointer/jitdb/pull/27

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,27 +20,16 @@ for doing EBT.
 To get the post messages of a specific author, you can do:
 
 ```js
-const { fromDB, query, and, type, author, toCallback } = require('jitdb/operators')
-
+const SecretStack = require('secret-stack')
 const caps = require('ssb-caps')
-const ssbKeys = require('ssb-keys')
-const path = require('path')
+const {query, and, type, author, toCallback} = require('ssb-db2/operators')
 
-const createSbot = require('secret-stack')({
-    caps: { shs: Buffer.from(caps.shs, 'base64') },
-})
-      .use(require("ssb-db2"))
-
-const dir = "./ssb"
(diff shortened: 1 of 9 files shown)


arj03:
I really like how this shortens the example code. Feels nice. My only concern is that now query here is different from the one in jitdb, a bit too magical. I know that most people won't use jitdb directly so won't notice, but still. What if instead of:

```
query(
  sbot.db,
  ...
```
it would be: 

```
sbot.db.query(
  ...
```

?

staltz:
@arj03 i'm fine with `sbot.db.query(`

arj03:
Like this right?

staltz:
oooops, I did a git push force and didn't realize you made a commit. i seemed to have deleted your commit, didn't even have time to read it :(

arj03:
Nevermind, they broke the tests ;-)

staltz:
ok both PRs should be in shape now

arj03:
Don't see the sbot.db.query change here?

staltz:
Should be here: https://github.com/ssb-ngi-pointer/ssb-db2/pull/10/files#diff-d7c9c4e37c37bdbf3fd7007e8e5a111ba1cbfecdd20b76f549a385ebc890e552R313 also see tests, it shows `db.query` but that's essentially the same `sbot.db.query`, just the sbot missing

arj03:
I can push the way I saw it to a separate branch, to compare?

staltz:
Yes

staltz:
Oh, the readme isn't updated. That's something that you can feel free to update in this branch or in master if you want. Right now I'm AFK

staltz:
Just to be sure, refresh this page

PR #13, opened 2020-11-14
Better createHistoryStream implementation and add tests

### LINKED ISSUE #6: Fix createHistoryStream to use new seq posibility

arj03:
This should fix #6


history-stream.js
```diff
@@ -16,17 +16,25 @@ exports.init = function (sbot, config) {
...
     const values = opts.values === false ? false : true
 
-    const query = {
-      type: 'EQUAL',
-      data: {
-        seek: sbot.db.jitdb.seekAuthor,
-        value: opts.id,
-        indexType: "author",
-        indexAll: true
+    let query = require("./operators").author(opts.id)
```
mixmix, line 23: Question :
 Are you making something with exactly same api? Seems like if it's going to be subdomained under ssb.db, changing the api there might be fine? 

 given id isn't an option, I would prefer `createHistoryStream(id, opts)`

Either way a check that id is a feedId could be important? I'm asking because intetested to know how and where others are checking inputs like this

arj03: Yeah it is meant to have the exactly same API as it would be used for legacy replication. I don't think it makes sense the implement the other streaming methods. I added a check + test for the feed thing, thanks :-)

arj03:
Peer reviewed the last part with staltz

Issue #4, opened 2020-11-13
Migrate between ssb-db and ssb-db2


arj03:
Add methods for migrating data from old flume
 - Code for this in jitdb repo

staltz:
- [x] Basic migration script
- [x] Support `live: true`
- [ ] Reuse the same async-flumelog instance instead of recreating it
- [ ] If new log already exists, don't log scan with `old: true`
- [ ] If new log already exists, log scan with `live: true`
- [ ] If old log still exists, refuse to allow db2.add and db2.publish

arj03:
Close as per #18

PR #18, opened 2020-11-16
migration plugin for ssb-db=>ssb-db2

staltz:
- [x] Copy and adapt the script from jitdb ('copy-json-to-bipf-async')
- [x] Tests
- [x] Support `live: true`
- [x] Reuse the same async-flumelog instance instead of recreating it
- [x] If new log already exists with N msgs, skip N msgs scanned from old log
- [x] Progress indicator (percentual) events for migration
- [x] If old log does not exist, migration plugin does not crash
- [x] If old log still exists, refuse to allow db2.add and db2.publish
- [x] Update README
- [ ] API to nuke the old log
  - Protect ourselves against forking when maintaining new log alongside old log
  - "force only one source of truth"
  - consider deleting the old log, or replacing it with broken data

### DIFF
--- a/migration.js
+++ b/migration.js
@@ -1,12 +1,18 @@
+const fs = require('fs')
 const path = require('path')
 const pull = require('pull-stream')
 const Notify = require('pull-notify')
 const FlumeLog = require('flumelog-offset')
 const bipf = require('bipf')
 const jsonCodec = require('flumecodec/json')
+const Obv = require('obv')
 const debug = require('debug')('ssb:db2:migration')
 
 const blockSize = 64 * 1024
+
+function getOldLogPath(config) {
+  return path.join(config.path, 'flume', 'log.offset')
+}
 
 function skip(count, onDone) {
   let skipped = 0
@@ -20,28 +26,34 @@ function skip(count, onDone) {
   })
 }
 
-function getOldLogStreams(sbot, config, live) {
-  const old = !live
+function makeFileExistsObv(filename) {
+  const obv = Obv()
+  obv.set(fs.existsSync(filename))
+  return obv
+}
+
+function getOldLogStreams(sbot, config) {
   if (sbot.createRawLogStream && sbot.createSequenceStream) {
-    const logStream = sbot.createRawLogStream({ old, live })
-    if (live) return [logStream]
+    const logStream = sbot.createRawLogStream({ old: true, live: false })
+    const logStreamLive = sbot.createRawLogStream({ old: false, live: true })
     const sizeStream = pull(
       sbot.createSequenceStream(),
       pull.filter((x) => x >= 0)
     )
-    return [logStream, sizeStream]
+    return [logStream, logStreamLive, sizeStream]
   } else {
-    const oldLogPath = path.join(config.path, 'flume', 'log.offset')
+    const oldLogPath = getOldLogPath(config)
     const oldLog = FlumeLog(oldLogPath, { blockSize, codec: jsonCodec })
-    const logStream = oldLog.stream({ seqs: true, old, live, codec: jsonCodec })
-    if (live) return [logStream]
+    const opts = { seqs: true, codec: jsonCodec }
+    const logStream = oldLog.stream({ old: true, live: false, ...opts })
+    const logStreamLive = oldLog.stream({ old: false, live: true, ...opts })
     const notify = Notify()
     oldLog.since(notify)
     const sizeStream = pull(
       notify.listen(),
       pull.filter((x) => x >= 0)
     )
-    return [logStream, sizeStream]
+    return [logStream, logStreamLive, sizeStream]
   }
 }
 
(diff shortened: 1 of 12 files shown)


migration.js
```diff
@@ -0,0 +1,49 @@
...
+      bipf.encode(data, buf, 0);
+      return buf;
+    }),
+    function sink(read) {
+      read(null, function next(err, data) {
+        if (err && err !== true) throw err;
+        if (err) {
+          debug('done');
+          if (config._db2migrationCB) newLog.onDrain(() => {
+            newLog.close(config._db2migrationCB)
```
mixmix, line 35: Opinion : as the final step move log.offset to log.offset.old

Motivation is to avoid forking feeds... But now I realize that wouldn't work completely. Maybe ssb-db would benefit from a pr which checks if flumelog.bipf before creating a new log.. M

staltz: Hmm! Very good point about forking! It seems to me that if both logs are in use, we have to have a strong guarantee that only one of them are written.

log.js
```diff
@@ -4,9 +4,9 @@ var path = require('path')
 
 module.exports = function (dir, config) {
   config = config || {}
-    
+
   var log = OffsetLog(
-    path.join(dir, 'log.bipf'),
+    path.join(dir, 'flumelog.bipf'),
```
arj03, line 9: Guess this sort of depends on #19. With #19 this shouldn't be needed?

migration.js
arj03: I think the documentation should be very explicit about this. There can only be 1 "master" log, so you have to write to the old log as long as that is still used. Maybe add a guard to add and publish that will detect if there is an active migration and abort.

staltz: Yes I thought about that same guard idea some minutes ago. We could check if the old log file exists and decline `add` and `publish` on db2 until it's deleted. Why a delete: because other apps or processes may write to the old log (have to assume that the currently running process doesn't have exclusive access to ~/.ssb)

log.js
staltz: Yes, I rebased this PR now with the latest changes from master

migration.js
mixmix: But with a delete if another app starts it will just make a new log. Likely no one will replicate it tho

staltz: There has to be some smart way of forcing just one source of truth, I'm not sure yet which method, but there should be _some method_ is my point

mixmix: I've got it! Leave a log.offset which is not a valid log file at all, is instead a picture of a pony.
Any app trying to load an offset log will instead get a pony which will clog its machinery.

Ugly but works

staltz:
I updated this branch with support for live queries. Check the check list in the original post, :star: indicates what was recently done.

Note that this PR builds on top of the `close` branch, which is PR #27

migration.js
staltz: gonna 'resolve' this because I made a TODO for it in the orig post, and this inline comment refers to disappeared code

staltz:
Migration progress events now supported, `sbot.on('ssb:db2:migration:progress', listener)`

staltz:
@arj03 I'm thinking of instead of a migration plugin, putting migration methods on `sbot.db.*`. or... putting migration methods on the migration plugin.

If all the logic runs ASAP on plugin init, then migration starts ASAP, and there are cases where we need the app to control when migration starts, so it would be convenient to have a `start()` method that begins the (processing intensive) migration. we could also have a config to enable automatic migration ASAP. this is similar to ssb-conn's scheduler start(): https://github.com/staltz/ssb-conn/#config see `autostart`.

Also, I may have found other methods that we may need, such as one method to forever kill the old `log.offset`, replacing it with a pony image. (in practice I think this should be a small file with some text like "MIGRATED_TO_ASYNC_FLUMELOG" and which guarantees that it will crash flumelog-offset when read) so in total these methods:

- `start()`
- `dangerouslyKillOldLog()` (inspiration: [dangerouslySetInnerHTML](https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml))
- `oldLogExists()` an Obv used only internally to guard `db.add()` and `db.publish()`

thoughts?

staltz:
@arj03 Also, I kind of need `db.js` to be a secret-stack plugin that gets `sbot` as an input, because if `db.js` is going to use `migration` plugin methods, and `migration` plugin needs `sbot` (e.g. for `emit`, and to use `ssb-db1` methods to read the log). Would it be cleaner to review a separate PR that only puts `sbot` support inside `db.js`?

arj03:
I think its fine to give db.js access to sbot in index.js. And in this PR. Then it can give access to migrate. I'm thinking migrate would live there. This could also have the logic for starting the migration on init, if as you say, it has this automigrate flag set. I think this would be cleaner for the add and publish methods as well that it can check the migrate plugin for access.

staltz:
Thanks! just to clarify:

> I'm thinking migrate would live there.

there where?

arj03:
sbot.db2.migrate

staltz:
Everything done now except the API to nuke the old log. I suppose we are not in an urgent need for that API, and it should be done rather carefully, so this PR as is is ready for review and merge.

migration.js
mixmix: Ok, as long as we don't lose these ponies, it took me ages to find them

arj03:
Review notes:
 - The fixtures module is amazing here. Really nice we can put that in and get a decent db to test on.
 - Also I really like debug, and that we start using it consistently in the stack, it helped track down the timeout thingy

Please check my commit that fixes the timeout. I'm going to try some more manual tests.

staltz:
Nice that you removed the setTimeouts! I'm keen to get that implicit onDrain inside db.query to simplify the API. But in another PR of course

arj03:
Did a bunch of tests migrating my full log. The migrate here is nice in that is both tests the actual migrate and also base indexing. Works perfectly killing it in the middle, it will start off where it was killed. I tested this against the raw copy from jitdb. 

Raw copy: 48s
Indexing: 26s
Migrate: 125s

So its a bit slower but it has progress and does a bunch of other stuff. So I'm pretty happy where with this ended up.

I pushed a small commit that shaves off 6 seconds. Its so small that I'll just go ahead and merge this. Feel free to revise.

Great work @staltz, it is one of those tasks that seems simple but really is more complicated than expected.

staltz:
Oh, good job testing it on a real big log, that didn't occur to me to do pre-merging.

> Raw copy: 48s
Indexing: 26s
Migrate: 125s

Seems like a quite huge difference, should we take a look at why it's that slow?

> I pushed a small commit that shaves off 6 seconds. Its so small that I'll just go ahead and merge this. Feel free to revise.

:+1: but I expect the tests to fail, since one of the tests checks for a specific amount of progress events. just have to update the tests, not a big deal.

arj03:
Right :facepalm: Will do a bit more testing to see where that extra time comes from.

arj03:
Actually removing base index brings the time down to 53s. So it could be write contention of log.bipf and base index.

staltz:
> Actually removing base index brings the time down to 53s. So it could be write contention of log.bipf and base index.

That rings true to me.

staltz:
`if (progressCalls < 100 || progressCalls++ % 1000 == 0) {`

smart one-liner throttler

arj03:
So I guess the only thing left here is to update #4 with the minor things left.

staltz:
Yeah I made issue #30 so we can close #4

PR #20, opened 2020-11-16
refactor usage of debug()

staltz:
First I noticed that we could use `debug` with namespaces such as `"ssb:db2:stuff"`, which is standard practice in [debug](https://www.npmjs.com/package/debug), allowing us to filter them in runtime like this: `DEBUG=ssb:* node myapp.js` or `DEBUG=ssb:db2:* node myappjs`.

Then, I refactored a bit so that we can get rid of one argument out of the 8 (many!) arguments to `Plugin()`.

### DIFF
--- a/indexes/base.js
+++ b/indexes/base.js
@@ -1,7 +1,6 @@
 const bipf = require('bipf')
 const pl = require('pull-level')
 const pull = require('pull-stream')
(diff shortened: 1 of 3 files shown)


arj03:
Nice one. Easy merge

PR #21, opened 2020-11-16
refactor: move lines around, use const/let

staltz:
These are simple refactors, mostly just for my sanity as I try to navigate the codebase, hope it's useful:

- `require`s always at the top, unless needed for browser differentiation
- `const` preferred if we know we won't re-assign the thing afterwards
- `let` preferred over `var` (var allows you to do the declaration twice, allowing some sneaky bugs pass through sometimes)
- variables used in scope declared at the top, inner functions declared below that (see `db.js` for an example)

This PR doesn't contain anything else interesting other than code shuffling.

### DIFF
--- a/db.js
+++ b/db.js
@@ -22,7 +22,26 @@ exports.init = function (dir, config) {
   //const contacts = fullIndex.contacts
   //const partial = Partial(dir)
 
-  var post = Obv()
+  const indexes = {
+    base: baseIndex
+  }
+  const post = Obv()
+  const hmac_key = null
+  let state = validate.initial()
+
+  // restore current state
+  baseIndex.getAllLatest((err, last) => {
+    // copy to so we avoid weirdness, because this object
+    // tracks the state coming in to the database.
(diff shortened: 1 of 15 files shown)


db.js
```diff
@@ -226,20 +227,16 @@ exports.init = function (dir, config) {
...
 
   function clearIndexes() {
-    for (var index in indexes)
-      indexes[index].remove(() => {})
-  }
-
-  var indexes = {
-    base: baseIndex
+    for (const indexName in indexes)
+      indexes[indexName].remove(() => {})
```
staltz, line 239: The diff above looks weird, here's what's actually going on:

- `var indexes = {....` moved to the top of this file
- `for (var index....` changed to `for (const indexName....`

arj03:
Was review this. Looks good except there are now merge conflicts. So once resolved, ok to merge.

staltz:
Conflicts resolved :heavy_check_mark:

arj03:
Nice

PR #25, opened 2020-11-17
Format code automatically with Prettier

staltz:
Like https://github.com/ssb-ngi-pointer/jitdb/pull/32

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,4 @@
+{
+  "semi": false,
+  "singleQuote": true
+}
--- a/compat/ebt.js
+++ b/compat/ebt.js
@@ -1,7 +1,7 @@
 exports.init = function (sbot, config) {
   sbot.post = sbot.db.post
   sbot.getAtSequence = sbot.db.getMessageFromAuthorSequence
-  sbot.getVectorClock = function(cb) {
+  sbot.getVectorClock = function (cb) {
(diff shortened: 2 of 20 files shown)


arj03:
There is something wierd here. In the jitdb PR it removes space after function so:

from
```
    read(null, function next (err, data) {
```
to

```
    read(null, function next(err, data) {
```

But this does the opposite? Like

  sbot.getVectorClock = function (cb) {

staltz:
When the function is named, it doesn't add a space. When it's anonymous, it adds a name. Just think that there's ways this exact same sequence of chars: `function `

arj03:
So we should name ALL the functions? :)

arj03:
I mean I *really* dislike the space there :-) Everything else is fine. Maybe not such a bad deal because naming function makes debugging easier :thinking:

staltz:
No no ... it's just a formatting style, adds a space in the anonymous case.

staltz:
Yeah, naming functions can be good for debugging (also in production). The code style will turn out to be unsurprising most of the times, and sometimes it will annoy (me or you), but the benefit of this is that we agree to not talk about code style, the machine does it for us, and the core benefit here is avoiding typing. (Nowadays I code sloppy style and let prettier format it for me)

arj03:
Yeah, merging. And now I can always spot when the function is not named, so that maybe a win in the end :-)

PR #26, opened 2020-11-17
Fix indexing performance by properly chunking writes

arj03:
This is rather embarrassing, I was fixing a bug related to unwrittenSeq a while ago and accidentily broke the chunkSize mechanism so it was more or less writing the whole time. With this commit indexing is 3x faster. 

I did discover some thing while tracking this down: chunksize 512 seems pretty good, I tried other values. Also the decode part is a large part of the indexing. The overhead of level is not so bad.

### DIFF
--- a/indexes/base.js
+++ b/indexes/base.js
@@ -78,7 +78,7 @@ module.exports = function (log, dir, feedId) {
       }
     }
 
-    if (batchBasic.length) return data.seq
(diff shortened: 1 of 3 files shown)


staltz:
I don't understand these parts perfectly yet, so I just trust you in this case, so merge away. :)

PR #27, opened 2020-11-18
add close() in db.js

### LINKED ISSUE #17: db2.close() needed

staltz:
fixes #17

### DIFF
--- a/db.js
+++ b/db.js
@@ -4,6 +4,7 @@ const validate = require('ssb-validate')
 const keys = require('ssb-keys')
 const path = require('path')
 const Obv = require('obv')
+const promisify = require('promisify-4loc')
 const jitdbOperators = require('jitdb/operators')
 const JITDb = require('jitdb')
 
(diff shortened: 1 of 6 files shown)


db.js
```diff
@@ -262,6 +263,16 @@ exports.init = function (dir, config) {
     })
   }
 
+  async function close(cb) {
```
staltz, line 266: promisify used here so we can run all these child `close` in parallel and use Promise.all to collect results. writing the equivalent with pure callbacks is a mind twist

arj03: wouldn't you be able to not have to use async on the function if you resolved the promise.all to the cb? Basically that is the only async call there.

staltz: oh yeah, that's another way of doing it

```js
  function close(cb) {
    const tasks = []
    tasks.push(promisify(log.close)())
    for (const indexName in indexes) {
      tasks.push(promisify(indexes[indexName].close)())
    }
    Promise.all(tasks).then(cb)
  }
```

the difference is that `async function` would allow using it as either `await db.close()` or `db.close(() => { ..... })`, the normal function wouldn't.

i'm fine either way

staltz: actually, `return Promise.all(tasks).then(cb)` then both usage styles are enabled!

staltz: Updated PR

arj03:
Perfecto. Thanks Andre

PR #29, opened 2020-11-19
remove unused code in the indexes

staltz:
I started by removing `query`, `fromDB`, `and` from the social index, and then realized there were other things not used. I'm pretty sure we are capable of bringing all these changes back, if in the future we realize we need them, but for now it seems better to not let dead code lying around possibly mistaken as important code.


arj03:
Looks good, would merge except for the merge conflicts.

staltz:
Conflicts resolved :heavy_check_mark:

Issue #32, opened 2020-11-23
Allow `migrate` to be used without the full ssb-db2


staltz:
ssb-db2 will immediately begin indexing the base index, we want to avoid that computational load. we want only migrate

PR #34, opened 2020-11-24
Allow `migrate` to be included alone

staltz:
For issue #32


arj03:
Looks good. Nice cleanup as well

PR #37, opened 2020-11-25
Support live queries in social index

arj03:
This depends on the [jitdb PR](https://github.com/ssb-ngi-pointer/jitdb/pull/38).

Allows the following kind of queries for all votes for a specific msg and all live updates:

```
db.query(
  and(votesFor(postMsg.key)),
  live(),
  toPullStream(),
  pull.drain((result) => {
 })
)
```

### DIFF
--- a/indexes/social.js
+++ b/indexes/social.js
@@ -2,7 +2,8 @@ const bipf = require('bipf')
 const pl = require('pull-level')
 const pull = require('pull-stream')
 const Plugin = require('./plugin')
-const { offsets } = require('../operators')
+const json = require('level-codec/lib/encodings').json
+const { offsets, liveOffsets } = require('../operators')
 
 // 3 indexes:
 // - root (msgId) => msg seqs
@@ -24,7 +25,7 @@ module.exports = function (log, dir) {
   let batch = []
 
   function writeData(cb) {
-    level.batch(batch, { keyEncoding: 'json' }, cb)
+    level.batch(batch, { keyEncoding: json }, cb)
     batch = []
   }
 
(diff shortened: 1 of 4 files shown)


arj03:
Test not working because of the dependency on jitdb. Will update this once that is merged.

indexes/social.js
```diff
@@ -2,7 +2,8 @@ const bipf = require('bipf')
 const pl = require('pull-level')
 const pull = require('pull-stream')
 const Plugin = require('./plugin')
-const { offsets } = require('../operators')
+const json = require('level-codec/lib/encodings').json
```
arj03, line 5: This was the problem @staltz ;-)

arj03: level understands 'json' fine but level-post does not :(

staltz: Aahhh I see now

arj03:
Should be ready for review now

indexes/social.js
```diff
@@ -102,54 +103,61 @@ module.exports = function (log, dir) {
...
+        if (live) {
+          const ps = pull(
+            pl.read(level, Object.assign({}, opts, { live, old: false })),
+            pull.map(parseInt10)
+          )
+          cb(null, liveOffsets(data.map(parseInt10), ps))
+        } else cb(null, offsets(data.map(parseInt10)))
+      })
+    )
+  }
```
staltz, line 120: Thought: I wonder if this could use `pull-cat` instead of being a nested callback looking thing.

Related to that, I wonder if `LIVEOFFSETS` could *only* have `ps`, and then we could do `and(offsets(data.map(parseInt10)), liveOffsets(ps))` and update the logic in JITDB so that LIVEOFFSETS is only about the pull-stream.

Just wondering of a refactor. This obviously only affects code style/quality, but arguably it's important so we don't end up with flumedb type of code quality.

arj03: I was thinking about that when doing the jitdb PR, it requires a bit finesse around liveOffsets for the normal queries, but could be done I guess.

staltz (approved):
This PR looks nice, small, and straight to the point, so I'd say lets merge it.

About the comment I made above, I made this issue: https://github.com/ssb-ngi-pointer/jitdb/issues/41

staltz:
Oh! One thing important before merging: put `level-codec` in package.json

staltz:
Can be done after merging too :)

arj03:
woops :)

staltz:
And! Instead of level-codec, just use flumecodec which is anyway a shortcut to level-codec, and we already have flumecodec imported

PR #40, opened 2020-11-27
Don't run live stream until after we finish write the seq

arj03:
One more of these *facepalm* bugs. With this, the index time of private (that I'm working on) goes down from 10 seconds to 2 seconds :-) And social from 12 to 7.

### DIFF
--- a/indexes/plugin.js
+++ b/indexes/plugin.js
@@ -61,8 +61,7 @@ module.exports = function (
       }
     }
 
-    function liveStream()
-    {
+    function liveStream() {
       isLive = true
       log.stream({ gt: seq.value, live: true }).pipe({
         paused: false,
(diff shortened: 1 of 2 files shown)


staltz:
Awesome gains! Can totally merge. Have one inline comment though

indexes/plugin.js
```diff
@@ -61,6 +61,15 @@ module.exports = function (
       }
     }
 
+    function liveStream()
+    {
```
staltz, line 65: I'm curious why didn't the code formatter git hook apply here? It should run automatically when you commit

arj03: Yeah that it strange, I just ran it manually and it made some minor changes. Will commit and merge.

arj03: oh, maybe its because it is only running on staged files? I just git commit -a

PR #41, opened 2020-11-27
fix migrate plugin handling of old log existence

staltz:
This was necessary to solve some race conditions I got when putting this plugin inside Manyverse.

Specifically, the flumelog-offset existed (as a file) but was of empty size, and doing a `live` streaming on that caused a crash in flumelog-offset.

This PR does two things:

- `fileExists` checks both that it exists and size > 0
- some exponential backoff retries of the `migrate.start()` function, handling corner cases such as closing the sbot

Tested that it works in production in Manyverse

### DIFF
--- a/migrate.js
+++ b/migrate.js
@@ -99,7 +99,7 @@ exports.init = function init(sbot, config, newLogMaybe) {
   let retryPeriod = 250
   let timer
 
-  function oldLogMissingRetry(fn) {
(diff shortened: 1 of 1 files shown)


arj03:
Fine by me

staltz:
@arj03 thanks! Can you also make a new release?

arj03:
@staltz sure, 0.2.0 released :)

PR #42, opened 2020-11-30
disable import of Partial

staltz:
`./indexes/partial` imports `../waiting-queue` which doesn't exist, and this causes `noderify` to crash when building the backend in Manyverse.

I'd need this merged and released soon, if you can, @arj03


arj03:
Ok, merging and releasing a new version

staltz:
wow that was fast

arj03:
Published as 0.3.0

PR #43, opened 2020-11-30
Remove pull-drain-gently

staltz:
Since `migrate` is I/O bound, not CPU bound, pull-drain-gently never gains us much, so I'd like to remove it to simplify the codebase. Should be easy to add back if we ever need it again. See also the other refactor commit contained here

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,10 +157,6 @@ Note that the `start` behavior is the same: you can either start it automaticall
 sbot.db2migrate.start()
 ```
 
-### Migrating gently
-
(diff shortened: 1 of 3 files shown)


staltz:
I'm puzzled with these tests not going anywhere. Locally they run just fine. In CI I can't even see the logs of whats going on

arj03:
Yeah that is strange, no debug is hard to debug :(

staltz:
Ah, so the `setInterval` is not being destroyed, so the tests hang forever. That's my guess. I'll do something about it, like a test timeout at least, so we detect these stuff in the future too.

staltz:
Updated this PR, tests *should* pass now

arj03:
Right, makes sense. Want a new release of this as well?

staltz:
@arj03 Rather a new release of #44

PR #44, opened 2020-11-30
add manifest for migrate plugin

staltz:
I needed this in order to call `start()` in Manyverse, new version would be appreciated

### DIFF
--- a/migrate.js
+++ b/migrate.js
@@ -83,6 +83,12 @@ function scanAndCount(pushstream, cb) {
 
 exports.name = 'db2migrate'
 
+exports.version = '0.3.0'
+
+exports.manifest = {
+  start: 'sync',
+}
+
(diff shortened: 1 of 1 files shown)


arj03:
I guess with the other fix, this should now be 0.4.0?

arj03:
Or better, include package.version

staltz:
> Or better, include package.version

actually, please not this, `noderify` would end up bundling the whole `package.json` into the final js file, and this time X modules using this pattern is more costly than 1 line of code (that rarely is updated)

staltz:
> I guess with the other fix, this should now be 0.4.0?

when making the commit for `0.4.0` you can also update this line of code in `migrate.js`

arj03:
0.4.0 released

Issue #3, opened 2020-11-13
Private groups


arj03:
Plan:
 - [x] add decryptor module that sits between log and views and keeps track of which messages can be decrypted and which ones are private
 - [x] make small wrapper around atomically for browser support
 - [x] add tests for private (indexes, get)
 - [x] Add EBT test
 - [x] be 100% that anything exposed on the network will not leak private messages. Test createHistoryStream & EBT. 
 - [x] Test [isPrivate](https://github.com/ssb-ngi-pointer/ssb-db2/blob/master/operators/index.js#L52)
 - [ ] extend ssb-fixtures with private groups https://github.com/ssb-ngi-pointer/ssb-fixtures/issues/1
 - [ ] add ability to decrypt private groups. This means extending the decryptor module to know about new keys and to be able to pass old messages that can now be decrypted to indexes. This is the hardest part because of concurrency of keys and log messages coming in. One option would be to simply pause the ordinary log while a new key is added. We need to figure out exactly how fast this is. The idea is to have a key store with a key + bit saying if the indexes know of the messages from this key or not. Because the indexes are not reduce, it doesn't matter if we crash in the middle and do some indexes multiple times.

arj03:
Been thinking about this. 

I think we need to store message encrypted as ssb-db. Then a crypto store that sits between the flumelog and views, including jitdb. It would have an 2 bitmap indexes: encrypted messages and messages we can decrypt.

On stream it would use the can decrypt index to only decrypt messages needed. This should be very efficient.

When added a new group to key to store it would run through the messages in can't decrypt yet (diff between the bitmaps) and update the "can decrypt" bitmap. The messages that can be decrypted this way would be given to the views. They would then either just update their indexes (in jitdb, base and social indexes this should be really fast (<1s)) or do a full reindex for reduce style. We need to consider this for about & contacts.

staltz:
Great idea to use bit vectors like this! In general I'm favorable of using bitvectors as a way of decreasing the search space for subsequent indexes.

arj03:
Thanks. I'm hoping that @mixmix might be interested in pairing on this as he has a lot of experience in this area already.

mixmix:
This sounds great. Originally I thought that we would want to reindex a smaller subset : encrypted + can't yet decrypt + author.

[...]

PR #46, opened 2020-12-02
Support private box1

arj03:
This PR adds support for decrypting private messages. This follows the same pattern as in ssb-db where we automatically decrypt messages and add metadata to message to know that they are private. This can then be used for querying, the flipside is that we need to be sure that we don't leak these private message over the network. Luckily the network "attached" code is much easier to spot here because the streams like historystream is not meant to be used internally. I have added tests for EBT and made sure that createhistorystream does not leak private messages in the tests. I also fixed the isPrivate operator and made a test for that. 

The way this is structured is that private is an index that sits between the log and indexes including jitdb. It keeps of: what messages are encrypted, what messages can be decrypted. As messages are streamed through this, the index will get updated and when reused in another index or a get, it uses the *can-be-decrypted* index will be used for fast decryption. On my log of 1.4 million messages, 112k are encrypted and of those I can decrypt 1300 messages. Because of this the overhead of decryption is more or less constant and independent of the number of indexes.

Some rough numbers from my laptop running on battery for creating indexes base + social index:
 - don't decrypt anything: 36 sec
 - decrypt: 50 sec
 - index base, stop and then reindex social: 50

So the overhead is roughly 1 second for 100k messages.

This should lay the foundation for adding support for decrypting private groups.


arj03:
Related issue #3

Would love your input on this @mixmix and of course @staltz :-)

arj03:
Random fact: the integer compression on what can be encrypted saves 50% of the size on that file which is important because it will be updated regularly while building the base index. This is also why the chunksize was changed fro 512 to 2048. Another note, atomically is also about 1 second faster than atomic file.

arj03:
Maybe @cryptix you will find some of this interesting as something to compare to the go code.

indexes/plugin.js
```diff
@@ -24,7 +24,7 @@ module.exports = function (
 
   const level = Level(indexPath)
   const META = '\x00'
-  const chunkSize = 512
+  const chunkSize = 2048
```
staltz, line 27: I'm not challenging anything, just want to ask what's the back story to this, and what did you learn

staltz: Nevermind, i just read your recent comment. :)

arj03:
Pushed a small commit for a theoretical problem where someone uses `get` on a seq that was not run through on index yet. This shouldn't happen because where would they get the seq, but anyway better safe than sorry.

mixmix:
Request - will this support me registering read keys for messages.

Eg there is a private thread about a security issue. We resolve it, then decide to disclose the thread to the community.

At the moment that looks like revealing the read keys for every message in the thread individually. A person cannot practically then read the thread (without worrying some script?). 

I would like to be able to later support a new message type, `disclose` which takes a bundle of keys and loads them into your db so that that while thread will now "just work".

This will require a key store, which I don't expect you to build. I just want to make sure your solution will support this (which I think it will? )

indexes/private.js
```diff
@@ -0,0 +1,148 @@
...
+          if (content) return reconstructMessage(data, content)
+        }
+      }
+    } else if (data.seq > latestSeq) {
+      if (streaming) latestSeq = data.seq
+
+      let p = 0 // note you pass in p!
+
+      p = bipf.seekKey(data.value, p, bValue)
+      if (~p) {
```
mixmix, line 112: Can we drop this ~
I would prefer more accessible code p >= 0

indexes/private.js
```diff
@@ -0,0 +1,148 @@
...
+        const pContent = bipf.seekKey(data.value, p, bContent)
+        if (~pContent) {
+          const type = bipf.getEncodedType(data.value, pContent)
+          if (type === StringType) {
+            encrypted.push(data.seq)
+
+            const content = ssbKeys.unbox(
+              bipf.decode(data.value, pContent),
+              keys
+            )
```
mixmix, line 122: Just checking - I don't want private box 1 baked into the implementation. Can you please make it an example implementation if an unboxer which is added and can later be extracted?

Also i think you should jump to making sure jump to making sure you can implement an. Unboxer for private groups. I found it a little tricky to implement and given private box was simple relatively. I think you would end up with a tidier setup if you made sure the complex case works

mixmix (commented):
Haven't had chance to read all. Just left couple comments

indexes/private.js
arj03: This is a good suggestion. I created this PR as the first step to solving #3. It doesn't go all the way we need to add the key store and the boxers concept. The main point of this was to get the pipeline in, tets and then we can add the other stuff.

arj03:
> Request - will this support me registering read keys for messages.

Yes, the mechanism for this should be very close to how I think private groups would work. I'm assuming you are talking about the unbox key for private box 1 here?

mixmix:
ssb-private1 and also ssb-tribes (DMs + group messages) have the concept of a shareable readKey per message.

We don't have to build support more, just leave space for it.
I wonder if there should be a db which maps msgkey to known decryption key. (not there are two sorts if decryption key, the msgKey which gives you access to everything , the readKey which gives access only to the content, no header data)

arj03:
Had a talk with Andre, going to merge this now. As said this is mostly a stepping stone.

mixmix:
All good. Keep up the great momentum

PR #47, opened 2020-12-03
Fix register index for queries

arj03:
While testing https://github.com/ssb-ngi-pointer/jitdb/pull/52, I ran into this problem when using the db.query and social indexes which makes me a bit sad. Should be introduce a self or how do you we best handle this @staltz ?

### DIFF
--- a/db.js
+++ b/db.js
@@ -268,7 +268,7 @@ exports.init = function (sbot, dir, config) {
 
     if (indexes[index.name]) throw 'Index already exists'
 
-    indexes[index.name] = index
+    this.indexes[index.name] = index
   }
 
   function onDrain(indexName, cb) {


staltz:
I understand the solution, but before we talk about alternative solutions, why was this problem occurring? I can't see the root cause

arj03:
I happens with this:

```
const sbot = SecretStack({ appKey: caps.shs })
  .use(require('./'))
  .use(require('./social')) // include index
  .call(null, {
    keys,
    path: dir,
  })

console.time("getting links")
sbot.db.query(
  and(votesFor('%hax.sha256')),
  toCallback((err, result) => {
    console.timeEnd("getting links")

    console.time("getting links2")
    sbot.db.query(
      and(votesFor('%DLNtxLj1bCyzU96vGkfgjcvf8WY4WlJgU7VL3EBvx08=.sha256')),
      toCallback((err, result) => {
        console.timeEnd("getting links2")
      })
    )

    console.log(err)
    console.log(result)
  })
)
```

staltz:
Do you remember what the error looked like? `indexes is undefined` or what?

arj03:
The problem is that indexes for query does not include social

staltz:
I'm gonna give this a try in a debugger and try to understand the bug. I fully trust you that this is a solution (by the way feel free to merge), i just want to really know what's going on

staltz:
Oh, so these two console.logs print out slightly different objects:

```diff
  function registerIndex(Index) {
    const index = Index(log, dir)

    if (indexes[index.name]) throw 'Index already exists'

    indexes[index.name] = index
+   console.log(indexes)
+   console.log(this.indexes)
  }
```

That's unfortunate, it means that the `indexes` object when attached to the returned object becomes a different object (does it get cloned?) from the one in the closure. I don't understand why are they different, why does it get "cloned"?

arj03:
I was hoping you would know :-) Maybe its one of those `this` is strange things?

staltz:
Oh I figured it out! secret-stack recursively goes through all deep fields of the returned object, and only stops going inside when it's a function, so it decides to go inside the `db.indexes` field, cloning every subfield.

Some alternative solutions: `db.getIndexes()` instead of `db.indexes`, or `self`

arj03:
nice debugging!

staltz:
I think the least buggy solution is getIndexes() because this way we are sure all cases refer to the same object. With `self`, secret-stack is still going to clone indexes and potentially pass it on to other APIs, and we want to avoid that entirely

PR #51, opened 2020-12-04
Convert key index in base to an jitdb prefix index

arj03:
This converts one of the level db indexes to a prefix index. I ran a test where I removed all indexes and lets it run followed by a query for a specific key. 

```
battery old: 53.2 s

arj@arj:~/dev/ssb-db2$ du --si ssb/db2/indexes/*
130M	ssb/db2/indexes/base
4,1k	ssb/db2/indexes/canDecrypt.index
529k	ssb/db2/indexes/encrypted.index
31M	ssb/db2/indexes/social

battery new: 51.7 s

arj@arj:~/dev/ssb-db2$ du --si ssb/db2/indexes/*
30M	ssb/db2/indexes/base
4,1k	ssb/db2/indexes/canDecrypt.index
529k	ssb/db2/indexes/encrypted.index
6,4M	ssb/db2/indexes/key.32prefix
6,4M	ssb/db2/indexes/offset.index
6,4M	ssb/db2/indexes/sequence.index
34M	ssb/db2/indexes/social
13M	ssb/db2/indexes/timestamp.index
```

While the speed up is not that big, one has to consider that the second had to create the jitdb indexes. The speedup will get bigger once we convert vote & root as well. But enough for today. 

The biggest difference is that this saves 94mb from the base index :)

Small note: running the same test on battery is around 32 seconds.

### DIFF
--- a/operators/index.js
+++ b/operators/index.js
@@ -5,6 +5,18 @@ function toBuffer(value) {
   return Buffer.isBuffer(value) ? value : Buffer.from(value)
 }
 
+function key(value) {
+  return {
+    type: 'EQUAL',
+    data: {
+      seek: seekers.seekKey,
+      value: Buffer.from(value),
+      indexType: 'key',
+      prefix: 32
+    },
+  }
+}
+
 function type(value) {
   return {
     type: 'EQUAL',
@@ -66,6 +78,7 @@ module.exports = Object.assign({}, jitdbOperators, {
   type,
   author,
   channel,
+  key,
   isRoot,
   isPrivate,
 })
--- a/db.js
+++ b/db.js
@@ -14,6 +14,8 @@ const Private = require('./indexes/private')
 const Migrate = require('./migrate')
 // const Partial = require('./indexes/partial')
 
+const { and, key, toCallback } = require('./operators')
+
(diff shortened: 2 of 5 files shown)


operators/index.js
```diff
@@ -5,6 +5,18 @@ function toBuffer(value) {
...
+  return {
+    type: 'EQUAL',
+    data: {
+      seek: seekers.seekKey,
+      value: Buffer.from(value),
+      indexType: 'key',
+      prefix: 32
+    },
+  }
+}
```
staltz, line 18: TODO in this file: convert the implementation of these to use `jitdbOperators.equal`, see also `type` and `author` below. Not strictly needed for this PR, though. The benefit is that we move away the details of the shape of the object to jitdbOperators.equal, because that's the function responsible for this

staltz (approved):
This is a really nice PR. feels good to delete code, and conceptually feels good to remove one huge leveldb index

operators/index.js
arj03: created an issue for that

arj03:
Yeah agree, I guess this space improvement should be very welcome on mobile. It will be really good in the browser at least.

Issue #52, opened 2020-12-04
Convert the implementation of these to use jitdbOperators.equal


arj03:
TODO in this file: convert the implementation of these to use `jitdbOperators.equal`, see also `type` and `author` below. Not strictly needed for this PR, though. The benefit is that we move away the details of the shape of the object to jitdbOperators.equal, because that's the function responsible for this

_Originally posted by @staltz in https://github.com/ssb-ngi-pointer/ssb-db2/pull/51#discussion_r536382954_

PR #53, opened 2020-12-05
refactor operators to use jitdb operators

staltz:
For issue #52 

Hardest part of this PR was setting up git config author and ssh keys on my phone


arj03:
Nice, net -44 lines :-)

Issue #54, opened 2020-12-06
Social plugin use prefix indexes


arj03:
Continuing #51. I'll keep this as an issue because I'm still exploring options here, wanted to share numbers.

Removing vote & root from index:

previous index time: 51.7s
new index time battery: 49.5s
social index goes from 34mb to 6.5mb.

Removing social index totally: 41.6s. Not a valid comparison, but just wanted the number.

Adding mentions to base index: 44s

So overhead of an extra log stream is 5.5s. This is probably the biggest takeaway.

PR #55, opened 2020-12-07
replace social.getMessagesByVoteLink() with a prefix index

staltz:
- [x] getMessagesByVoteLink
- [x] getMessagesByRoot
- [x] isRoot

Just wanted to push this branch so it's ready for testing.


arj03:
I think you can do root at the same time. Also for root you can see if we can't remove the isRoot function with the root prefix index. Maybe a test for that.

arj03:
With both you should get numbers similar to what I posted in #54.

staltz:
:+1: I'll add another commit for that

staltz:
getMessagesByRoot replaced :heavy_check_mark:

arj03:
@staltz this one I was thinking of: https://github.com/ssb-ngi-pointer/ssb-db2/blob/30f7bf61fde23eabc8d0c2312fe44afb2bf42a42/operators/index.js#L28

arj03:
@staltz create a jitdb 0.25.0 release for this :)

staltz:
This PR is ready for review

operators/index.js
```diff
@@ -6,31 +6,57 @@ const {
...
-  return equal(seekType, value, { indexType: 'value_content_type' })
+  return equal(seekType, value, {
+    indexType: 'value_content_type',
+  })
 }
 
 function author(value) {
-  return equal(seekAuthor, value, { indexType: 'value_author' })
+  return equal(seekAuthor, value, {
+    indexType: 'value_author',
```
staltz, line 28: I'm curious what would the performance look like if we made all these prefix indexes! `author`, `type` and `channel`.

That's something to benchmark soon

arj03: Yes, I think it def. makes sense for channels. Author would be interesting to contrast with using a deferred on the the level index to see. For type if we can get the sorted lookup in there then yes, otherwise I'm leaning towards a specific index for that because its so heavily used.

arj03:
This looks great. 

I'll try adding a query on load that makes sure these prefix indexes are updated so you only take 1 hit instead of 3.

PR #56, opened 2020-12-07
refactor indexType names

staltz:
Just a refactor, not too important, and requires a reindex, but I figured that more descriptive/explicit is better for our future sanity and other developers, and this is not yet used in production, so now is the time to do this kind of refactor.

E.g. imagine if in the future we add another kind of of `private` e.g. `value.content.private`, then it makes it very clear what is what by making it explicit from day one.


arj03:
Yep, makes sense. Lets fix that now instead of later ;)

Issue #35, opened 2020-11-24
Occasional levelup error `ReadError: Database is not open`


staltz:
I get this error randomly when running `DEBUG=* npm test`:

```
  refuses to db2.add() while old log exists

  secret-stack Adding transport net +316ms
  secret-stack Adding transform shs +1ms
  jitdb indexes path /tmp/ssb-db2-migrate/db2/indexes +20ms
  secret-stack net plugin creating transport host=undefined port=NaN scope=device,local,public +331ms
  secret-stack creating server net shs host=undefined port=NaN scope=device,local,public +14ms
  secret-stack net plugin creating transport host=undefined port=NaN scope=device,local,public +1ms
  secret-stack net plugin creating transport host=undefined port=NaN scope=undefined +0ms
  multiserver:net Listening on false:49915 +19ms
  async-flumelog opened file, since: 10447 +46ms
  async-flumelog getting offset 0 from disc +1ms
  ssb:db2:migrate new log has 11 msgs, will continue migrating +315ms
  ssb:db2:migrate reading old log +1ms
  ssb:db2:base got index status: { version: 1, seq: 7607, processed: 11 } +0ms
    ✔ should be truthy
    ✔ should be falsy
    ✔ error message is about the old log
  ssb:db2:migrate done migrating 0 msgs from old log +14ms
  async-flumelog getting offset 7607 from cache +22ms
  ssb:db2:base index scan time: 1ms, items: 12 +15ms

/home/staltz/ssb-db2/indexes/plugin.js:43
          if (err) throw err
                   ^
ReadError: Database is not open
    at maybeError (/home/staltz/ssb-db2/node_modules/.pnpm/levelup@4.4.0/node_modules/levelup/lib/levelup.js:339:32)
    at LevelUP.put (/home/staltz/ssb-db2/node_modules/.pnpm/levelup@4.4.0/node_modules/levelup/lib/levelup.js:211:7)
    at writeBatch (/home/staltz/ssb-db2/indexes/plugin.js:38:13)
    at Object.end (/home/staltz/ssb-db2/indexes/plugin.js:69:11)
    at Stream.abort (/home/staltz/ssb-db2/node_modules/.pnpm/async-flumelog@1.0.8/node_modules/async-flumelog/stream.js:150:15)
    at blocks.getBlock (/home/staltz/ssb-db2/node_modules/.pnpm/async-flumelog@1.0.8/node_modules/async-flumelog/stream.js:134:12)
[...]

PR #61, opened 2020-12-08
Use one log stream for all indexes

arj03:
The idea here is that instead of each plugin doing their own log stream we can have one stream driving them all. The improvement is substantial as was known from db1 and documented in #54. And the gain will only grow as new indexes are added. An added benefit is that with this we can merge #58 because in the case of a crash where base is then ahead of private, this will make sure that we continue from where private stopped. The overhead of calling the calls for indexes that are already up to date is minimal. Also because the indexes are pretty fast already, adding a new overhead will just give a 5 seconds delay before the normal indexes will start received new data.

### DIFF
--- a/db.js
+++ b/db.js
@@ -266,14 +266,12 @@ exports.init = function (sbot, dir, config) {
     const start = Date.now()
 
     const indexesRun = Object.values(indexes)
-    let isLive = false
 
     function liveStream() {
       debug('live streaming changes')
-      isLive = true
       log.stream({ gt: indexes['base'].seq.value, live: true }).pipe({
         paused: false,
-        write: (data) => indexesRun.forEach((x) => x.onData(data, isLive)),
+        write: (data) => indexesRun.forEach((x) => x.onData(data, true)),
       })
     }
 
@@ -284,12 +282,9 @@ exports.init = function (sbot, dir, config) {
 
     log.stream({ gt: lowestSeq }).pipe({
       paused: false,
-      write: (data) => indexesRun.forEach((x) => x.onData(data, isLive)),
+      write: (data) => indexesRun.forEach((x) => x.onData(data, false)),
       end: () => {
-        const tasks = []
-        indexesRun.forEach((index) => {
-          tasks.push(promisify(index.writeBatch)())
-        })
+        const tasks = indexesRun.map((index) => promisify(index.writeBatch)())
         Promise.all(tasks).then(liveStream)
 
         debug(`index scan time: ${Date.now() - start}ms`)
--- a/db.js
+++ b/db.js
@@ -66,7 +66,7 @@ exports.init = function (sbot, dir, config) {
   }
 
   function get(id, cb) {
-    onIndexesReady(() => {
+    onIndexesStateLoaded(() => {
       query(
         and(key(id)),
         toCallback((err, results) => {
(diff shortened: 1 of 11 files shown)


db.js
```diff
@@ -63,14 +66,16 @@ exports.init = function (sbot, dir, config) {
   }
 
   function get(id, cb) {
```
staltz, line 68: is this the only API that needs to be guarded with `onIndexesReady`?

db.js
```diff
@@ -280,7 +321,31 @@ exports.init = function (sbot, dir, config) {
...
+  function checkIndexesReady() {
+    let allOk = private.latestSeq.value !== undefined
+    debug(`private seq: ${private.latestSeq.value}`)
+    for (var index in indexes) {
+      if (indexes[index].seq.value === undefined) allOk = false
+      debug(`${index} seq: ${indexes[index].seq.value}`)
+    }
+
+    if (allOk) indexesReady.set(true)
+    else if (!closing) setTimeout(checkIndexesReady, 250)
```
staltz, line 337: Instead of this setTimeout, what if we set `indexesReady` in `updateIndexes()` when the log.stream ends?

db.js
```diff
@@ -256,6 +262,41 @@ exports.init = function (sbot, dir, config) {
...
+    debug(`lowest seq for all indexes ${lowestSeq}`)
+
+    log.stream({ gt: lowestSeq }).pipe({
+      paused: false,
+      write: (data) => indexesRun.forEach((x) => x.onData(data, isLive)),
+      end: () => {
+        const tasks = []
+        indexesRun.forEach((index) => {
+          tasks.push(promisify(index.writeBatch)())
+        })
```
staltz, line 292: It would be better to avoid mutation if we can make the code simpler:
```js
const tasks = indexesRun.map(index => promisify(index.writeBatch)())
Promise.all(tasks).then(liveStream)
```

db.js
```diff
@@ -256,6 +262,41 @@ exports.init = function (sbot, dir, config) {
...
+
+    const indexesRun = Object.values(indexes)
+    let isLive = false
+
+    function liveStream() {
+      debug('live streaming changes')
+      isLive = true
+      log.stream({ gt: indexes['base'].seq.value, live: true }).pipe({
+        paused: false,
+        write: (data) => indexesRun.forEach((x) => x.onData(data, isLive)),
```
staltz, line 276: I was looking back and forth at `isLive`, since it's defined up there, mutated here, used down there, etc, and I think now that I understand this code, it would have been easier to understand the code without this variable.

Just call `x.onData(data, true)` or `x.onData(data, false)` here and there, then we can avoid the variable and its mutation and mentally following how it's updated.

indexes/base.js
```diff
@@ -48,7 +47,9 @@ module.exports = function (log, dir, private) {
     batchJson = []
   }
 
-  function handleData(data) {
+  function handleData(data, processed) {
+    if (data.seq < seq.value) return
```
staltz, line 51: :+1:

staltz (changes requested):
Looking good! The reshuffle wasn't that complex after all. I like the `onData`/`writeBatch` pattern.

There's just one important question, and then a couple of code refactor suggestions. If you wish and you're busy with other stuff, I can do the refactors for you.

db.js
arj03: Thanks

db.js
arj03: Yeah this setTimeout is unfortunate. What it does here is that it needs to make sure that all indexes have loaded their state before really starting, this includes any indexes that was added after db was init'ed. So we can't really to it in updateIndexes. Its the other way around.

db.js
arj03: It would be better in query, but that api is sync

staltz: We could use a `deferred` to create a wait. Also, could we combine `onDrain` and `onIndexesReady` and have just one way of waiting? Does that make sense?

arj03: Deferred would be a good idea. I think those are two different things. onIndexesReady is a boot thing whole the other is for when data is added later. Maybe a better name would be onIndexStateLoaded instead of onIndexesReady?

db.js
arj03: Ah maybe 'multiserver:listening' can be used

arj03:
This should close #35 as well

arj03:
Pair programmed the last part.

PR #66, opened 2020-12-09
improve private index save/load methods

staltz:
What this PR does: see the inline comments

### DIFF
--- a/indexes/private.js
+++ b/indexes/private.js
@@ -20,15 +20,13 @@ module.exports = function (dir, keys) {
   const encryptedFile = path.join(indexesPath(dir), 'encrypted.index')
   const canDecryptFile = path.join(indexesPath(dir), 'canDecrypt.index')
 
-  function save(filename, seq, arr, cb) {
+  function save(filename, arr) {
     const buf = toBuffer(fic.compress(arr))
     const b = Buffer.alloc(4 + buf.length)
-    b.writeInt32LE(seq, 0)
+    b.writeInt32LE(latestSeq.value, 0)
     buf.copy(b, 4)
 
     writeFile(filename, b, { fsyncWait: false })
-      .then(() => cb())
-      .catch((err) => cb(err))
   }
 
   function load(filename, cb) {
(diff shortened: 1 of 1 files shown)


indexes/private.js
```diff
@@ -20,15 +20,13 @@ module.exports = function (dir, keys) {
   const encryptedFile = path.join(indexesPath(dir), 'encrypted.index')
   const canDecryptFile = path.join(indexesPath(dir), 'canDecrypt.index')
 
-  function save(filename, seq, arr, cb) {
```
staltz, line 23: The `seq` is always the `latestSeq` variable, so I just put that one directly inside here since it has access to that variable anyway.

I removed the cb, because we don't care about the results. If they succeed or fail, we don't care. Except maybe one day we want to put a `debug` for the error case, but that can be done locally, without putting the error in a `cb`

indexes/private.js
```diff
@@ -20,15 +20,13 @@ module.exports = function (dir, keys) {
...
+  function save(filename, arr) {
     const buf = toBuffer(fic.compress(arr))
     const b = Buffer.alloc(4 + buf.length)
-    b.writeInt32LE(seq, 0)
```
staltz, line 26: It seems we were writing an Obv in the buffer! that seems wrong, right? I made it obv.value instead. does this fix some weird bugs?

indexes/private.js
```diff
@@ -39,14 +37,16 @@ module.exports = function (dir, keys) {
...
   }
 
   function loadIndexes(cb) {
     load(encryptedFile, (err, data) => {
       if (err) {
         latestSeq.set(-1)
-        return cb(err)
+        if (err.code === 'ENOENT') cb()
+        else cb(err)
+        return
```
staltz, line 49: We care about errors, unless the index does not exist previously (which is a common case when we're starting up the database)

indexes/private.js
```diff
@@ -70,21 +70,22 @@ module.exports = function (dir, keys) {
...
-        save(encryptedFile, latestSeq, encrypted, () => {})
-        save(canDecryptFile, latestSeq, canDecrypt, () => {})
+        save(encryptedFile, encrypted)
+        save(canDecryptFile, canDecrypt)
       }, 1000)
     }
     cb()
   }
 
-  const bKey = Buffer.from('key')
```
staltz, line 87: this wasn't used

indexes/private.js
arj03: Oh 🤦, I was changing that variable back and forth and in shared log PR and didn't get it properly fixed everywhere it seeems. Nice catch.

arj03:
Thanks! :)

PR #67, opened 2020-12-09
replace setTimeout in onIndexesStateLoaded with promises

staltz:
`p-defer` is a ~6 LOC dependency. The approach here is to introduce a `stateLoaded` deferred promise (think: an Obv that can only be `set` once) instead of piggybacking on `seq`, and then collect all `stateLoaded` for all indexes using Promise.all

### DIFF
--- a/db.js
+++ b/db.js
@@ -318,31 +318,22 @@ exports.init = function (sbot, dir, config) {
     })
   }
 
-  let indexesStateLoaded = Obv()
-  indexesStateLoaded.once(updateIndexes)
-  let closing = false
-
-  function checkIndexesStateLoaded() {
-    let allOk = private.latestSeq.value !== undefined
-    debug(`private seq: ${private.latestSeq.value}`)
-    for (var index in indexes) {
-      if (indexes[index].seq.value === undefined) allOk = false
-      debug(`${index} seq: ${indexes[index].seq.value}`)
-    }
-
-    if (allOk) indexesStateLoaded.set(true)
-    else if (!closing) setTimeout(checkIndexesStateLoaded, 250)
-  }
-
-  checkIndexesStateLoaded()
-
   function onIndexesStateLoaded(cb) {
-    if (indexesStateLoaded.value === true) cb()
-    else indexesStateLoaded.once(cb)
+    if (!onIndexesStateLoaded.promise) {
+      const stateLoadedPromises = [private.stateLoaded]
+      for (var index in indexes) {
+        stateLoadedPromises.push(indexes[index].stateLoaded)
+      }
+      onIndexesStateLoaded.promise = Promise.all(stateLoadedPromises)
+    }
+    onIndexesStateLoaded.promise.then(cb)
   }
+
+  setTimeout(() => {
+    onIndexesStateLoaded(updateIndexes)
+  })
 
   function close(cb) {
-    closing = true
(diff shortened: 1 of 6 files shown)


db.js
```diff
@@ -318,31 +318,22 @@ exports.init = function (sbot, dir, config) {
...
+    onIndexesStateLoaded.promise.then(cb)
   }
 
-  checkIndexesStateLoaded()
-
-  function onIndexesStateLoaded(cb) {
-    if (indexesStateLoaded.value === true) cb()
-    else indexesStateLoaded.once(cb)
-  }
+  setTimeout(() => {
```
staltz, line 332: Without this, I got an error that I don't understand:

```
Error: Not opened
    at Request._openAndNotClosed (/home/staltz/oss/manyverse/ssb-db2/node_modules/.pnpm/random-access-storage@1.4.1/node_modules/random-access-storage/index.js:179:53)
    at Request._run (/home/staltz/oss/manyverse/ssb-db2/node_modules/.pnpm/random-access-storage@1.4.1/node_modules/random-access-storage/index.js:203:16)
    at RandomAccessFile.RandomAccess.run (/home/staltz/oss/manyverse/ssb-db2/node_modules/.pnpm/random-access-storage@1.4.1/node_modules/random-access-storage/index.js:111:12)
    at RandomAccessFile.RandomAccess.read (/home/staltz/oss/manyverse/ssb-db2/node_modules/.pnpm/random-access-storage@1.4.1/node_modules/random-access-storage/index.js:56:8)
    at Object.getBlock (/home/staltz/oss/manyverse/ssb-db2/node_modules/.pnpm/async-flumelog@1.0.10/node_modules/async-flumelog/index.js:99:11)
    at Stream._resume (/home/staltz/oss/manyverse/ssb-db2/node_modules/.pnpm/async-flumelog@1.0.10/node_modules/async-flumelog/stream.js:127:15)
    at Stream._next (/home/staltz/oss/manyverse/ssb-db2/node_modules/.pnpm/looper@4.0.0/node_modules/looper/index.js:11:9)
    at Stream.resume (/home/staltz/oss/manyverse/ssb-db2/node_modules/.pnpm/async-flumelog@1.0.10/node_modules/async-flumelog/stream.js:144:8)
    at Stream.pipe (/home/staltz/oss/manyverse/ssb-db2/node_modules/.pnpm/push-stream@11.0.0/node_modules/push-stream/pipe.js:19:25)
    at Stream.s.pipe (/home/staltz/oss/manyverse/ssb-db2/log.js:44:14)
```

anyway it made sense to put this setTimeout, to give time for other indexes to be registered, and maybe other things in secret-stack.

arj03: Hmm yeah, there is something during load. Also without this setTimeout you probably would not be waiting for any extra indexes like mentions.

arj03:
I like this promise pattern, just wish obv would support it out of the box so we didn't need the extra variables.

This now has a conflict in private. Can you fix that? I'm fine merging this.

staltz:
conflicts resolved :heavy_check_mark: 

> just wish obv would support it out of the box so we didn't need the extra variables.

yeah, obv doesn't have something like `Promise.all`, and including `multicb` into this mix could complicate things, so I tried the simplest pattern I could find for this problem. gladly Promise.all is built in and not a dependency

PR #69, opened 2020-12-09
use official leveldb isClosed() API

staltz:
The official API for checking if level is closed is https://github.com/Level/level#dbisclosed

I ran the tests 100 times and couldn't bump into #35 anymore. Still not sure if this is a definitive solution though

```
for i in $(seq 100); do npm test; sleep 1; done
```


arj03:
Much cleaner. Fingers crossed this fixes the problem for good ;-)

Issue #68, opened 2020-12-09
(Sometimes) errors when testing getAllLatest


staltz:
In the `master` branch as of now, this happens probabilistically. Might be related to the `stateLoaded` changes I made, since I haven't seen this one before.

```
  get all latest

    ✔ no err
    ✔ authors

    ✖ should be strictly equal
    ---------------------------
      operator: equal
      expected: '%vXvTSDTQ1DhKvrLGyb1gSrJ4I0XYCgFQKf9xtVWcrik=.sha256'
      actual:   '%Gav5Mr9yyb1B7nsbYQb6OwSnH4c5ddjy+eKpf/4LhGI=.sha256'
      at: db.getAllLatest (/home/staltz/oss/manyverse/ssb-db2/test/base-index.js:134:11)
      stack: |-
  Error: should be strictly equal
  at Test.assert [as _assert] (/home/staltz/oss/manyverse/ssb-db2/node_modules/.pnpm/tape@5.0.1/node_modules/tape/lib/test.js:260:54)
  at Test.bound [as _assert] (/home/staltz/oss/manyverse/ssb-db2/node_modules/.pnpm/tape@5.0.1/node_modules/tape/lib/test.js:84:32)
  at Test.strictEqual (/home/staltz/oss/manyverse/ssb-db2/node_modules/.pnpm/tape@5.0.1/node_modules/tape/lib/test.js:424:10)
  at Test.bound [as equal] (/home/staltz/oss/manyverse/ssb-db2/node_modules/.pnpm/tape@5.0.1/node_modules/tape/lib/test.js:84:32)
  at db.getAllLatest (/home/staltz/oss/manyverse/ssb-db2/test/base-index.js:134:11)
  at pull.collect (/home/staltz/oss/manyverse/ssb-db2/indexes/base.js:108:9)
  at /home/staltz/oss/manyverse/ssb-db2/node_modules/.pnpm/pull-stream@3.6.14/node_modules/pull-stream/sinks/reduce.js:10:5
  at /home/staltz/oss/manyverse/ssb-db2/node_modules/.pnpm/pull-stream@3.6.14/node_modules/pull-stream/sinks/drain.js:20:24
  at drain (/home/staltz/oss/manyverse/ssb-db2/node_modules/.pnpm/stream-to-pull-stream@1.7.3/node_modules/stream-to-pull-stream/index.js:126:18)
[...]

PR #70, opened 2020-12-09
remove unused log arg to plugins

staltz:
After the log-sharing PR, this argument is no longer needed


staltz:
There were some probabilistic errors occurring in CI, they were #68 and #71

arj03:
Hmm yeah, we need to get to the bottom of those random errors. PR looks good.

PR #72, opened 2020-12-09
bug the debug

arj03:
Show me non-deterministic universe

### DIFF
--- a/db.js
+++ b/db.js
@@ -304,10 +304,15 @@ exports.init = function (sbot, dir, config) {
         const index = indexes[indexName]
         if (!index) return cb('Unknown index:' + indexName)
 
+        debug(`drain got log: ${index.seq.value}, index: ${index.seq.value}`)
+
         if (index.seq.value === log.since.value) {
           cb()
         } else {
           const remove = index.seq(() => {
+            debug(
+              `drain seq update got log: ${index.seq.value}, index: ${index.seq.value}`
+            )
             if (index.seq.value === log.since.value) {
               remove()
               cb()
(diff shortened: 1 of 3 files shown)


arj03:
Hello there buggy bear, nice to see you in the flesh

arj03:
It seems as if the log is actually in sync, the failing test has the same log diff as the previous ones that are ok. 

```
2020-12-10T07:26:27.282Z ssb:db2 drain got log: 301429, index: 301429
2020-12-10T07:26:27.536Z ssb:db2 drain got log: 302644, index: 302644, log diff: 1215
2020-12-10T07:26:27.793Z ssb:db2 drain got log: 303859, index: 303859, log diff: 1215
2020-12-10T07:26:28.050Z ssb:db2 drain got log: 305074, index: 305074, log diff: 1215
```

New theory, searching for a specific key using prefix index can sometime return the wrong result.

arj03:
Added some more debug locally, it is actually an error from level

```
NotFoundError: Key not found in database [["@t9HfCmbvcKjjlsMBi/eTxzlACc462uHKJwiN+GFVfLQ=.ed25519",700]]
  at /home/arj/dev/ssb-db2/node_modules/levelup/lib/levelup.js:188:15
  at /home/arj/dev/ssb-db2/node_modules/encoding-down/index.js:75:21
```

arj03:
Haha, think I know now

arj03:
Seems to fix it, will just run it 1 more time after this finishes to make sure that it looks good. I could also reproduce it locally. @staltz do you think we should keep the basic test, seems good for spotting these kind of things but will take a bit more time to run.

staltz:
Good solution! About the test, it used to be 30s now it's about 2min, I'm not sure if that's bad or acceptable. Is the test deterministically failing for the previous implementation or is it probabilistically failing? If it's probabilistically, then seems we could reduce the number from 1000 to ~10 or so, then the probability of a failure is lower, but is still > 0.

arj03:
Right, I will lower it a bit, then one can always increase the number if you want to stress test more.

arj03:
Will do a proper PR

PR #83, opened 2020-12-15
fix migrate for ssb-db@19.3.1

staltz:
Manyverse uses ssb-db@19.3.1 and the `sizeStream` in migrate wasn't working, while ssb-db2 uses ssb-db@20.3.0 for the tests.

I changed ssb-db to 19.3.1 in this module's package.json, then tests didn't pass either. Then I fixed the implementation until tests passed for 19.3.1. And then I changed package.json back to 20.3.0. Unfortunately I don't know a way of installing both versions and running tests, so we just have to remember to manually test 19.3.1 if we change `migrate`'s implementation. But now we know `migrate` works for both versions.

### DIFF
--- a/migrate.js
+++ b/migrate.js
@@ -32,29 +32,41 @@ function makeFileExistsObv(filename) {
 }
 
 function getOldLogStreams(sbot, config) {
-  if (sbot.createRawLogStream && sbot.createSequenceStream) {
+  if (sbot.createRawLogStream && sbot.status) {
     const logStream = sbot.createRawLogStream({ old: true, live: false })
     const logStreamLive = sbot.createRawLogStream({ old: false, live: true })
-    const sizeStream = pull(
-      sbot.createSequenceStream(),
-      pull.filter((x) => x >= 0)
-    )
-    return [logStream, logStreamLive, sizeStream]
+    const getSize = () => sbot.status().sync.since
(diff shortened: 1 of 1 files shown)


migrate.js
```diff
@@ -32,29 +32,41 @@ function makeFileExistsObv(filename) {
...
+      raw: true,
+      old: true,
+      live: false,
+    })
+    const logStreamLive = sbot.createLogStream({
+      raw: true,
+      old: false,
+      live: true,
+    })
+    const getSize = () => sbot.status().sync.since
```
arj03, line 52: Should this be since.value instead?

arj03: There same a few lines up

arj03:
Just a small comments, otherwise looks good.

migrate.js
staltz: It's weird, isn't it? But it's correct,

- see 20.3.0: https://github.com/ssbc/ssb-db/blob/da89b7552e054c9e8baaad78506714e13a62a7ec/index.js#L107
- see 19.3.1: https://github.com/ssbc/ssb-db/blob/6c0cdd580b6166b106283e3f1745585e189252ce/index.js#L94

arj03: OMG!

arj03:
Looks good :-)

PR #84, opened 2020-12-15
Fix delete

### LINKED ISSUE #60: Fix delete methods
Currently the delete methods are broken because of major refactors. We should make them work again and add tests ;)

arj03:
Fixes #60 by making delete and deleteFeed work again. This depends on https://github.com/ssb-ngi-pointer/jitdb/pull/74 and https://github.com/flumedb/async-flumelog/pull/13.

### DIFF
--- a/db.js
+++ b/db.js
@@ -14,7 +14,14 @@ const BaseIndex = require('./indexes/base')
 const Private = require('./indexes/private')
 // const Partial = require('./indexes/partial')
 
-const { and, fromDB, key, deferred, toCallback } = require('./operators')
+const {
+  and,
+  fromDB,
+  key,
+  author,
+  deferred,
+  toCallback,
+} = require('./operators')
 
 function getId(msg) {
   return '%' + hash(JSON.stringify(msg, null, 2))
@@ -141,16 +148,15 @@ exports.init = function (sbot, config) {
     })
   }
 
-  function del(key, cb) {
+  function del(msgId, cb) {
     const guard = guardAgainstDuplicateLogs('del()')
     if (guard) return cb(guard)
 
-    // FIXME: this doesn't work anymore after changing base index
-    baseIndex.keyToSeq(key, (err, seq) => {
+    jitdb.all(key(msgId), 0, false, true, (err, results) => {
       if (err) return cb(err)
-      if (seq == null) return cb(new Error('seq is null!'))
+      if (results.length === 0) return cb(new Error('seq is null!'))
 
-      log.del(seq, cb)
+      log.del(results[0], cb)
     })
   }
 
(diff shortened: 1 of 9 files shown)


log.js
```diff
@@ -36,8 +36,7 @@ module.exports = function (dir, config, private) {
   let originalGet = log.get
   log.get = function (seq, cb) {
     originalGet(seq, (err, res) => {
-      if (err && err.code === 'flumelog:deleted') cb()
```
arj03, line 39: This was removed because we already handle this in jitdb and the only other place this is used is in EBT index and actually if that happens then this means that you delete a message from an author but not the whole feed and this scenario shouldn't be supported in EBT anyway.

arj03:
While writing this I noticed that we really need proper documentation of the different methods. I think after merging this in would be a good point to add that.

db.js
```diff
@@ -114,49 +121,35 @@ exports.init = function (sbot, dir, config) {
...
-          )
-        }
+    jitdb.all(author(feedId), 0, false, true, (err, results) => {
+      push(
+        push.values(results),
+        push.asyncMap((seq, cb) => {
+          log.del(seq, cb)
+        }),
+        push.collect((err) => {
+          if (!err) {
```
staltz, line 147: This seems a bit backwards and made me stop and read it carefully. I think if it were written in a more standard way, would make it easy to skim:

```js
if (err) cb(err)
else {

}
```

db.js
```diff
@@ -114,49 +121,35 @@ exports.init = function (sbot, dir, config) {
...
+  function del(msgId, cb) {
     const guard = guardAgainstDuplicateLogs('del()')
     if (guard) return cb(guard)
 
-    // FIXME: this doesn't work anymore after changing base index
-    baseIndex.keyToSeq(key, (err, seq) => {
+    jitdb.all(key(msgId), 0, false, true, (err, results) => {
       if (err) return cb(err)
-      if (seq == null) return cb(new Error('seq is null!'))
+      if (results.length === 0) return cb(new Error('seq is null!'))
```
staltz, line 130: The error message seems not relevant anymore, how about `'cannot delete ' + msgId + ' because it was not found'`

staltz (changes requested):

test/delete.js
```diff
@@ -0,0 +1,58 @@
...
+const db = sbot.db
+
+test('index, delete, query', (t) => {
+  const post = { type: 'post', text: 'Testing!' }
+  const post2 = { type: 'post', text: 'Testing 2!' }
+
+  db.publish(post, (err, postMsg) => {
+    t.error(err, 'no err')
+
+    db.log.onDrain(() => {
```
staltz, line 36: Just a heads up that this might need to be rebased with #85 so that it's `db.getLog()` etc

staltz:
What happened to the tests? They ran for 5h then were cancelled

arj03:
Wierd, not sure. I plan on looking at this again and rebasing tonight or tomorrow.

arj03:
Found the problem. It was a missing cb.

arj03:
This should be ready now. @staltz do you want a release of master before getting this in?

arj03:
Awww naawww, an error :(

arj03:
I'm really fond of these automated tests, found a bug that didn't show up locally. The problem (in a test) is also in master.

staltz:
> This should be ready now. @staltz do you want a release of master before getting this in?

Actually yeah, that would be useful. :)

This one can be merged IMO

arj03:
Great, released current master as 0.8.0

PR #86, opened 2020-12-15
add contact(feedid) operator

staltz:
This one is just because I want to replace another simple API inside Manyverse and seemed easy to do.

### DIFF
--- a/operators/index.js
+++ b/operators/index.js
@@ -9,8 +9,9 @@ const {
   seekVoteLink,
   seekMentions,
   pluckLink,
+  seekContact,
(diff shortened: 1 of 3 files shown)


arj03:
Easy merge, what are you using it for out of curiosity?

staltz:
@arj03 I would use it to load a contact messages (latest old and then live) authored by A about B, both A and B fixed feedIds, and checking whether the contact msg is encrypted. This is to supplement the lack of "isPrivatelyFollowing" / "isPrivatelyBlocking" APIs in ssb-friends.

PR #92, opened 2020-12-20
Better documentation

arj03:
I wanted to add documentation to our methods and started looking at add. I never really used the method in ssb-browser only internally, but noticed that in ssb-db it validates. I didn't want to change for no reason, so moved some things around a bit and made it so that `add` works like in ssb-db. I still want to have a look at #7, but for now this should be good enough.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,15 +65,21 @@ sbot.db.query(
 
 ### Extra plugins
 
-An extra index plugin that is commonly needed in SSB communities is the **full-mentions** index. It has one method:
+An extra index plugin that is commonly needed in SSB communities is
+the **full-mentions** index. It has one method: getMessagesByMention.
 
- - getMessagesByMention
+Although this accomplishes the same as the previous `mentions()`
+example, this plugin is meant as an example for application developers
+to write their own plugins if the functionality of JITDB is not
+enough. JITDB is good for indexing specific values, like
+`mentions(alice.id)` which gets its own dedidated JITDB index for
+`alice.id`. But when querying mentions of several feeds or several
+messages, this creates many indexes, so a specialized index makes more
+sense.
 
-Although this accomplishes the same as the previous `mentions()` example, this plugin is meant as an example for application developers to write
-their own plugins if the functionality of JITDB is not enough. JITDB
-is good for indexing specific values, like `mentions(alice.id)` will gets its own dedidated JITDB index for `alice.id`. But when querying mentions of several feeds or several messages, this would creates many indexes, so a specialized index makes more sense.
-
-What `full-mentions` does is index all possible mentionable items at once, using Leveldb instead of JITDB. You can include it and use it like this:
+What `full-mentions` does is index all possible mentionable items at
+once, using Leveldb instead of JITDB. You can include it and use it
+like this:
 
 ```js
 const SecretStack = require('secret-stack')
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -160,27 +173,81 @@ const sbot = SecretStack({appKey: caps.shs})
...
 
-Because ssb-db2 also begins indexing basic metadata once it's included as a plugin, this may cost more (precious) CPU time. **If you are not yet using db2 APIs** but would like to migrate the log anyway, in preparation for later activating db2, then you can include only the migration plugin, like this:
+Because ssb-db2 also begins indexing basic metadata once it's included
+as a plugin, this may cost more (precious) CPU time. **If you are not
+yet using db2 APIs** but would like to migrate the log anyway, in
+preparation for later activating db2, then you can include only the
+migration plugin, like this:
 
 ```js
 const sbot = SecretStack({appKey: caps.shs})
```
mixmix, line 193: I don't like using appKey. I think we should use `{ caps: { shs } }` generally as this (to me) feels clearer. Especially when there's caps.sign (which could be confused with appKey... shs is more of a networkKey...)

mixmix (changes requested):

README.md
arj03: Thanks, makes sense.

db.js
```diff
@@ -113,7 +113,7 @@ exports.init = function (sbot, config) {
     )
   }
 
-  function add(msg, cb) {
+  function rawAdd(msg, cb) {
     const guard = guardAgainstDuplicateLogs('add()')
```
staltz, line 117: Because `rawAdd` is an internal method, it doesn't make sense to have `guardAgainstDuplicateLogs` anymore, could you remove it? :)

README.md
```diff
@@ -29,9 +29,9 @@ const SecretStack = require('secret-stack')
 const caps = require('ssb-caps')
 const {and, type, author, toCallback} = require('ssb-db2/operators')
 
-const sbot = SecretStack({appKey: caps.shs})
+const sbot = SecretStack({ caps })
```
staltz, line 32: Does this work like this? (I know mix commented already)
I often use `SecretStack({appKey: require('ssb-caps').shs})` but I would **love** to use `SecretStack({ caps: require('ssb-caps') })`. Basically just asking if you're sure it works

README.md
```diff
@@ -138,49 +147,107 @@ const config = {
...
+this method unless you know exactly what you are doing.
+
+### deleteFeed(feedId, cb)
+
+Delete all messages of a specific feedId. Compared to `del` this
+method is safe to use.
+
+### publish(msg, cb)
+
+Convinience method for validating and adding a message to the database
```
staltz, line 226: Convenience

README.md
```diff
@@ -138,49 +147,107 @@ const config = {
...
+written by the feed running the secret-stack.
+
+### add(msg, cb)
+
+Validate and add a raw message (without id and timestamp) to the
```
staltz, line 231: Should we say "message value" as in `msg.value` instead? Maybe it helps clarify what this object should look like?

db.js
```diff
@@ -37,8 +37,8 @@ exports.manifest = {
   publish: 'async',
   del: 'async',
   deleteFeed: 'async',
-  validateAndAdd: 'async',
-  validateAndAddOOO: 'async',
+  addOOO: 'async',
+  addOOOStrictOrder: 'async',
```
staltz, line 41: Good rename

staltz (changes requested):

README.md
arj03: That's how ssb-server does it https://github.com/ssbc/ssb-server/blob/53c29aa819ad82b068c0b8e25db664215d1b2734/index.js#L6

db.js
arj03: Thanks, yes!

README.md
staltz: TIL

README.md
arj03: That word is not convinient to use ;-)

arj03:
Thanks for the review. Pushed an update.

staltz (approved):

PR #93, opened 2020-12-23
rename offset to seq and vice-versa

staltz:
This should come after https://github.com/ssb-ngi-pointer/jitdb/pull/78 is merged and published, then we need to update this one to use the new version of jitdb.

It's here as a draft to prove that tests pass.


indexes/plugin.js
```diff
@@ -44,17 +44,17 @@ module.exports = function (
...
     } else cb()
   }
 
-  function onData(data, isLive) {
+  function onData(record, isLive) {
     // maybe check if for us!
-    let unwritten = handleData(data, processed)
-    if (unwritten > 0) unWrittenSeq = data.seq
+    let unwritten = handleData(record, processed)
+    if (unwritten > 0) unWrittenOffset = record.seq
```
arj03, line 57: this should be offset

arj03: Ah right, a bit confusing

compat/history-stream.js
```diff
@@ -17,14 +17,14 @@ exports.permissions = {
 exports.init = function (sbot, config) {
   sbot.createHistoryStream = function (opts) {
     // default values
-    const seq = opts.sequence || opts.seq || 0
+    const sequence = opts.sequence || opts.seq || 0
```
arj03, line 20: This is a nice benefit of the new naming, much easier to spot these things.

indexes/plugin.js
arj03: Pushed a comment instead, I wonder if we should land a change to async-flumelog before merging this. It would cut down some of the confusion and it seems to be quite a small change. What do you think @staltz?

staltz: Sounds good by me. Actually async-flumelog is a bit of a wildcard in the flumedb family. It's not quite conforming to flumedb, which would require the name seq. I'd be more comfortable if async-flumelog wouldn't have flume in the name, but even if it has flume in the name, I'm fine with renaming seq to offset there

arj03: Yeah, one could always do a wrapper to make it conform to flume seq again. I think I'll do that (figure out a good name) and move the repo to ssb-ngi-pointer then.

arj03:
Converted this to use async-log. Now this depends on https://github.com/ssb-ngi-pointer/jitdb/pull/79 which depends on https://github.com/ssb-ngi-pointer/async-log/pull/14

indexes/plugin.js
staltz: Nice to move it to this GH org :+1:

arj03:
Should be ready to go now :)

PR #94, opened 2020-12-27
Minor tweaks to index building

arj03:
Just some minor things I noticed

### DIFF
--- a/indexes/base.js
+++ b/indexes/base.js
@@ -40,8 +40,8 @@ module.exports = function (log, dir, private) {
   }
 
   function handleData(record, processed) {
-    if (record.offset < offset.value) return
-    if (!record.value) return // deleted
+    if (record.offset < offset.value) return batch.length
+    if (!record.value) return batch.length // deleted
 
     let p = 0 // note you pass in p!
     const pKey = bipf.seekKey(record.value, p, bKey)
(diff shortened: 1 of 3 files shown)


indexes/base.js
```diff
@@ -40,8 +40,8 @@ module.exports = function (log, dir, private) {
   }
 
   function handleData(record, processed) {
-    if (record.offset < offset.value) return
-    if (!record.value) return // deleted
+    if (record.offset < offset.value) return batch.length
```
arj03, line 43: this is better when running in live mode

indexes/full-mentions.js
```diff
@@ -55,7 +55,7 @@ module.exports = function (log, dir) {
               ) {
                 batch.push({
                   type: 'put',
-                  key: [mention.link, 'm', shortKey],
```
arj03, line 58: smaller indexes, since we now only store 1 here, no need to add the extra m

indexes/base.js
```diff
@@ -40,8 +40,8 @@ module.exports = function (log, dir, private) {
...
+    if (!record.value) return batch.length // deleted
 
     let p = 0 // note you pass in p!
     const pKey = bipf.seekKey(record.value, p, bKey)
```
staltz, line 47: Another tiny suggestion: `const buffer = record.value` since `record.value` is read many times, we save some `.` object lookups. Probably just a tiny boost in performance, but it doesn't hurt readability, and this is a hot path for performance.

staltz (approved):

indexes/base.js
arj03: It is indeed a tiny bit faster. Thanks.

Issue #95, opened 2020-12-28
Migrate takes a long time before copying messages


staltz:
I just debugged here on my Manyverse why is there a multi-second delay between pressing 'like' and seeing the like appear (multi-second means more than 20 seconds). It seems logical that the `vote` is published on the old log, but db2-migrate has to move that to the new log before allowing db2 queries to succeed.

Turns out that the initial `scanAndCount` inside db2 migrate takes 31 seconds to complete. I think there's two bad ideas going on:

- `scanAndCount` scans through the *entire* new log just to count how many records it has (`msgCountNewLog`)
- After that, the *entire* old log is scanned, but its first `msgCountNewLog` messages are dropped due to the use of a `skip` pull-stream operator

I think we need a way of setting the cursor on the old log directly. I'm thinking that we need to save the offset on the old log where we know we have migrated to the new log. Maybe we need to save this tiny piece of information (let's call it `migratedOffset`) in a separate file? It would also be good to have some protection and sanity checks against overriding or duplicating messages on the new log in case `migratedOffset` happens to refer to be out of sync with the actual migration status (this could happen e.g. if the new log is saved and a crash happens right before `migratedOffset` is saved, whereas in the default case they should be saved "at the same time").

Thoughts @arj03 ?

arj03:
Yeah this is a bit of a state nightmare. Especially because you really don't want to add the same message twice and adding a add guard that checks if the message is already there is way to slow.

What if instead of tried finding the last message in the new log first. It should be possible to start close to the end and get the last one. After finding that, you should be able to get the seq (in old flume terminology ;-)) of that in the old log. Then it should be a matter of stream from there?

staltz:
Yeah you're right about the state nightmare, I'll try your suggestion. :)

staltz:
Help me out a bit? Given a msg key, how do I retrieve its (byte) offset position in the old log? Or even better, with ssb-db(1) APIs, since that might be the only interface I have with the old log. I can't quite find an API for this in ssb-db.

arj03:
It should be using the [key index](https://github.com/ssbc/ssb-db/blob/2162df147d513e774f2fb3d582ff8055558a31e7/create.js#L69).

staltz:
Do you know a way of accessing that from the public API? They don't seem to be exported

staltz:
I have an idea. It's semi-hacky, you'll see

PR #97, opened 2020-12-28
improve performance of migrate plugin

staltz:
For issue #95. I tested this in my production Manyverse and it gave a significant improvement. Time between "press like" and "like appears" dropped from (approximately) 45 seconds to (approximately) 2 seconds.

This PR has a somewhat hacky solution. It basically expects `sbot.get(msgkey, cb)` to call the `cb` with 3 arguments, `cb(err, msg, offset)`. If the 3rd argument is undefined, migrate plugin fallsback to what it was doing before: scan the new log, counting the records, then scan the old log.

If `sbot.get` is patched to provide the offset in the third arg, this improves performance a lot. Otherwise, it's the same. I intend to ship a hacky `ssb-db` to provide the third arg. We can work in the `ssb-db` repo to add this support, or add another API, either way, we can rather easily get rid of the hacky idea by improving `ssb-db` APIs.

### DIFF
--- a/migrate.js
+++ b/migrate.js
@@ -7,18 +7,7 @@ const jsonCodec = require('flumecodec/json')
 const Obv = require('obv')
 const debug = require('debug')('ssb:db2:migrate')
 const { BLOCK_SIZE, oldLogPath, newLogPath } = require('./defaults')
-
-function skip(count, onDone) {
-  let skipped = 0
-  return pull.filter((x) => {
-    if (skipped >= count) return true
-    else {
-      skipped++
-      if (skipped === count && onDone) onDone(x)
-      return false
-    }
-  })
-}
+const seekers = require('./seekers')
 
 function fileExists(filename) {
   return fs.existsSync(filename) && fs.statSync(filename).size > 0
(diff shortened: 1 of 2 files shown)


test/migration.js
```diff
@@ -79,7 +79,56 @@ test('migrate moves msgs from old log to new log', (t) => {
   )
 })
 
-test('migrate keeps new log synced with old log being updated', (t) => {
+test('migrate moves msgs from ssb-db to new log', (t) => {
```
staltz, line 82: This test is the same as the previous, but it installs `ssb-db`. This is to make sure that some code paths are covered (maybe someday we should add a code coverage tool).

migrate.js
```diff
@@ -91,6 +82,71 @@ function scanAndCount(pushstream, cb) {
...
+function findMigratedOffset(sbot, oldLog, newLog, cb) {
+  if (!sbot.get) {
+    debug('running in inefficient mode because no ssb-db is installed')
+    inefficientFindMigratedOffset(newLog, oldLog, cb)
+    return
+  }
+
+  newLog.onDrain(() => {
+    if (typeof newLog.since.value !== 'number' || newLog.since.value < 0) {
+      cb(null, -1)
```
staltz, line 120: This `if` condition will be triggered by the new test added. Without that new test, we were potentially exposed to the lack of this `if` condition passing all tests, when in production it could break.

arj03:
I looked over the changes. Looks good from what I can see. One question about the delay, this would only be if you like right after startup? Because I would assume that once everything is in sync and its just doing live sync, then things should be faster?

arj03 (approved):

staltz:
>  One question about the delay, this would only be if you like right after startup? Because I would assume that once everything is in sync and its just doing live sync, then things should be faster?

Yes, exactly, I startup the app and immediately press like. But it's still a significant thing to fix because it's very likely that the user will press like within the first 45 seconds of using the app.

arj03:
Indeed, good to know and glad you found a solution to this :)

PR #98, opened 2020-12-28
refactor bipf pointers and other minor refactors

staltz:
Just some shuffling around of code, and some renames. I was studying the code, trying to understand how `private` affects performance (I have some ideas I'd like to discuss with you later, because it seems `private.decrypt` during log stream has a noticeable impact on initial sync), but while studying code I noticed these possible refactors. Feel free to ask me to rollback some if they've gone too far.

### DIFF
--- a/indexes/private.js
+++ b/indexes/private.js
@@ -12,7 +12,7 @@ const Debug = require('debug')
 const { indexesPath } = require('../defaults')
 
 module.exports = function (dir, keys) {
-  let latestOffset = Obv()
+  const latestOffset = Obv()
   const stateLoaded = DeferredPromise()
   let encrypted = []
   let canDecrypt = []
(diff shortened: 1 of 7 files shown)


indexes/private.js
```diff
@@ -110,46 +106,43 @@ module.exports = function (dir, keys) {
     return { offset: record.offset, value: buf }
   }
 
+  const bValue = Buffer.from('value')
+  const bContent = Buffer.from('content')
+
   function decrypt(record, streaming) {
```
staltz, line 112: This diff might be hard to read, but essentially it's just flattening the nested `if`s. Plus, `record.value` => `recBuffer`.

arj03:
Looked over the changes. Looked good to me. Thanks

arj03:
>  it seems private.decrypt during log stream has a noticeable impact on initial sync)

Yes it does. Box1 is not the fastest.

PR #101, opened 2020-12-30
Use obz instead of obv

arj03:
I ran into [this](https://github.com/ssbc/obz/commit/9f8e7de8d6e6f13b08b7b76ff68a440220a72096) problem.

Also sneaked in a small helper function that makes the code a tiny bit nicer to use.


staltz:
Could `getIndex` be used in the full mentions operator?

arj03:
Right, thanks.

staltz (approved):
Looks good!

PR #103, opened 2021-01-01
Faster add and better addOOOStrictOrder

### LINKED ISSUE #102: Add it waaaay too slow
I'm testing initial sync and found that [this](https://github.com/ssb-ngi-pointer/ssb-db2/blob/e97403ba1ca7b996edde304d6a29ab000b26cc56/db.js#L123) is way too slow. The comment above is actually also wrong now that we use jitdb for this. The overhead per message is something like 250ms, so its not going to work. I'm thinking for add in order we might skip this check, but I don't like that we had it before (when using base index) and it was ok.

arj03:
This should fix #102 and #7.

ssb-validate knows the state of each feed, so if you add the same message twice or a message not in order it will error. We can rely on this and only check if message already exists in the strict OOO case.

I also reworked addOOOStrictOrder so that it doesn't touch the internal validate state.

### DIFF
--- a/db.js
+++ b/db.js
@@ -110,20 +110,17 @@ exports.init = function (sbot, config) {
     )
   }
 
-  /**
-   * Beware:
-   * There is a race condition here if you add the same message quickly
-   * after another because baseIndex is lazy. The default js SSB
-   * implementation adds messages in order, so it doesn't really have
-   * this problem.
-   */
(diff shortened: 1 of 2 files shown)


staltz (approved):
Reviewed lightly, on mobile

arj03:
Thanks :)

PR #105, opened 2021-01-02
Add isPublic operator

arj03:
To complement isPrivate

### DIFF
--- a/operators/index.js
+++ b/operators/index.js
@@ -84,7 +84,7 @@ function isPrivate() {
   return equal(seekPrivate, bTrue, { indexType: 'meta_private' })
 }
 
-function isNotPrivate() {
+function isPublic() {
   return equal(seekMeta, undefined, { indexType: 'meta' })
 }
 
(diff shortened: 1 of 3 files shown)


operators/index.js
```diff
@@ -83,6 +84,10 @@ function isPrivate() {
   return equal(seekPrivate, bTrue, { indexType: 'meta_private' })
 }
 
+function isNotPrivate() {
+  return equal(seekMeta, undefined, { indexType: 'meta_private' })
```
staltz, line 88: Shouldn't the indexType be `'meta'`?

staltz (changes requested):

operators/index.js
arj03: Sure, I guess that is a bit cleaner

test/operators.js
```diff
@@ -10,6 +10,7 @@ const {
   and,
   type,
   isPrivate,
+  isNotPrivate,
```
mixmix, line 13: Why not isPublic?

arj03: A very good suggestion :D Thanks

staltz (approved):
Thanks! `isPublic` is good too

PR #107, opened 2021-01-02
add benchmark scripts (in CI as well)

staltz:
Some of this was copied from jitdb. I'm interested in adding this now, before I attempt one trick to speed up initial indexing.

### DIFF
--- a/benchmark/index.js
+++ b/benchmark/index.js
@@ -22,8 +22,8 @@ if (!skipCreate) {
   fs.appendFileSync(reportPath, '## Benchmark results\n\n')
 
   const SEED = 'sloop'
-  const MESSAGES = 10000
-  const AUTHORS = 200
+  const MESSAGES = 100000
+  const AUTHORS = 2000
 
   test('generate fixture with flumelog-offset', (t) => {
     generateFixture({
--- a/benchmark/index.js
+++ b/benchmark/index.js
@@ -1,0 +1,77 @@
+const test = require('tape')
+const fs = require('fs')
+const path = require('path')
+const rimraf = require('rimraf')
+const mkdirp = require('mkdirp')
+const generateFixture = require('ssb-fixtures')
+const SecretStack = require('secret-stack')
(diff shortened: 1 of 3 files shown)


arj03:
Looks OK. I noticed the scripts in jitdb have a high variance, might be good to run these things with something that averages over multiple runs. Can't remember the package now.

staltz:
Yeah, in CI we don't need to worry too much about time, it can take even many minutes to complete. 

Note I used the same fixture size for both JITDB and ssb-db2. This means that the durations in JITDB are around half a second each test, but here the initial indexing is 9 seconds, it gives you a sense of scale that initial (base index) indexing (with decrypt) is many times heavier than JITDB indexing.

arj03 (approved):

staltz:
@arj03 I made a new commit, can you review it?

benchmark/index.js
```diff
@@ -42,16 +44,44 @@ if (!skipCreate) {
...
+
+  pull(
+    fromEvent('ssb:db2:migrate:progress', sbot),
+    pull.filter((progress) => progress === 1),
+    pull.take(1),
+    pull.drain(() => {
+      const duration = Date.now() - start
+      t.pass(`duration: ${duration}ms`)
+      fs.appendFileSync(reportPath, `- Migration: ${duration}ms\n`)
+      // wait for new log FS writes to finalize
```
arj03, line 70: Shouldn't a onDrain work here?

arj03:
It's good to see this split into migrate and indexing

benchmark/index.js
staltz: Oh, going to try that.

staltz: Oh, the problem here is that I don't have `sbot.db`, so I don't have access to the `log` :(

arj03: Ah right, okay.

arj03: Feel free to merge :)

PR #109, opened 2021-01-03
benchmark: dont migrate during initial indexing

staltz:
We're using the "inefficient mode" of migrate plugin because https://github.com/ssbc/ssb-db/pull/319 is not yet deployed, so the `initial indexing` benchmark is disturbed by migrate running in the background (it's not actually migrating anything, it's just scanning the entire old log). Once we have efficient mode for migrate, then we *could* keep this line of code in, but anyway it doesn't make sense to keep it in.

### DIFF
--- a/benchmark/index.js
+++ b/benchmark/index.js
@@ -83,8 +83,6 @@ test('initial indexing', (t) => {
     .use(require('../'))
     .call(null, { keys, path: dir })
 
-  sbot.db2migrate.start()
-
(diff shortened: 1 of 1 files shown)


staltz:
Compare the bot's comment above with this: https://github.com/ssb-ngi-pointer/ssb-db2/pull/107#issuecomment-753616636

> - Migration: 3195ms
> - Initial indexing: 4046ms

PR #111, opened 2021-01-04
Always keep track of latest offset

arj03:
As it is now, when the normal log finishes processing, all indexes will get a [write](https://github.com/ssb-ngi-pointer/ssb-db2/blob/05a7779953c22fae7846c05e113d5fb16b022c28/db.js#L263). If an index doesn't write something for each message, lets say its only interested in a particular type, then the last write might not write the final offset which means that you can't do something like onDrain on that index. Furthermore when you restart the app, it has to go through all those messages again it just checked. In the live case we have the same problem, but here we don't want to write for each message so we debounce the write a bit.

### DIFF
--- a/indexes/plugin.js
+++ b/indexes/plugin.js
@@ -54,7 +54,8 @@ module.exports = function (
   function onData(record, isLive) {
     // maybe check if for us!
     const unwritten = handleData(record, processed)
-    if (unwritten > 0) unWrittenOffset = record.offset
+    if (isLive && unwritten > 0) unWrittenOffset = record.offset
+    else unWrittenOffset = record.offset
     processed++
 
     if (unwritten > chunkSize || isLive) writeBatch(() => {})
--- a/indexes/plugin.js
+++ b/indexes/plugin.js
@@ -1,5 +1,6 @@
 const Obv = require('obz')
 const Level = require('level')
+const debounce = require('lodash.debounce')
 const path = require('path')
 const Debug = require('debug')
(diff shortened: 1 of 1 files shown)


arj03:
I just did a full sync in the browser with this and I don't have any problems with the browser getting slow after initial sync, so it seems like this is the last part I need in order to be able to merge in the ssb-db2 changes :-)

arj03:
@staltz I'm not sure if you are using indexes in manyverse yet. But the problem I was experiencing was that if the indexes catches up with the log, then they will go into live mode and before this would write for every message :(

staltz:
Good catch! I'm not using these indexes yet in Manyverse, it's useful you fixed this at this point.

indexes/plugin.js
```diff
@@ -51,13 +52,15 @@ module.exports = function (
     } else cb()
   }
 
+  const liveWriteBatch = debounce(writeBatch, 250)
+
   function onData(record, isLive) {
-    // maybe check if for us!
     const unwritten = handleData(record, processed)
-    if (unwritten > 0) unWrittenOffset = record.offset
+    unWrittenOffset = record.offset
```
staltz, line 59: Why was the if removed?

arj03: That is this part: If an index doesn't write something for each message

staltz: Ah I see

staltz (approved):

PR #114, opened 2021-01-05
Release 0.17.0

staltz:
Sent as a PR just to make sure we don't concurrently do the same thing.

Also updates AAOL


arj03:
Hehe :100:  Was thinking of doing a release once the tests works :-)

PR #120, opened 2021-01-15
Add getMsg

### LINKED ISSUE #119: db.get() should return full msg
when db.get() succeeds, it calls the cb with `results[0].value`. I would need this to return msg so that it has both `msg.key` and `msg.value`, not `msgValue`.

arj03:
This should fix #119. Please note it also changes get to err if the key is not found. This is the same behaviour as ssb-db.

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,7 +206,11 @@ sbot.db2migrate.start()
 
 ### get(msgId, cb)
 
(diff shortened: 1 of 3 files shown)


staltz (approved):
Looks good!

PR #130, opened 2021-01-19
Fix migrate to not read old log pre-decrypted

staltz:
1st commit is failing test, 2nd commit fix

### DIFF
--- a/migrate.js
+++ b/migrate.js
@@ -68,13 +68,10 @@ function guardAgainstDecryptedMsg(msg) {
     (msg && msg.meta && msg.meta.private) ||
     (msg && msg.value && msg.value.meta && msg.value.meta.private)
   ) {
-    console.error(
+    return new Error(
       'ssb:db2:migrate was about to write ' +
         'private message *decrypted* to disk'
     )
-    return true
-  } else {
-    return false
   }
 }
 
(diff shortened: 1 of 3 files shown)


staltz:
The commit above **should fail** the test but it's not failing. Help @arj03

staltz:
I have some idea what it could be, be right back.

staltz:
Oh I think the test passes for ssb-db@20, because `createRawLogStream` actually will stream messages in encrypted format, it's just ssb-db@19 that doesn't have this API and that's exactly the version of ssb-db that Manyverse still uses.

staltz:
Okay, now what this PR does is install `ssb-db@19.3.1` (what Manyverse uses) and all tests pass. I'm curious what will `@github-actions` bot say about perf benchmarks.

arj03:
nice this is quite a bit faster :) I checked that the oldlog obz returns the encrypted message as well. Just to be 100% sure. The flume log stream is nicely easy to reason about in that regard.

arj03 (approved):

staltz:
@arj03 Thank you for taking a look. I added one more commit, just to be double sure that the obz doesn't return a private message decrypted: https://github.com/ssb-ngi-pointer/ssb-db2/pull/130/commits/fe62d228422e5be63e765de51b79fdc0653e06c0 This module is tricky, I'd rather play double safe than sorry.

arj03:
I would rather have guardAgainstDecryptedMsg hard fail if that happens because your new log will be screwed in that you might have skipped messages.

arj03:
Good to go now :)

Issue #133, opened 2021-01-20
Only live


arj03:
I actually thought this query would work:

```
pull(
  sbot.db.query(
    live({ old: true }),
    descending(),
    paginate(1000),
    toPullStream()
  ),
  pull.drain(msg => {
    console.log(msg)
  })
)
```
but it fails with:

```
/home/arj/dev/ssb-db2/node_modules/jitdb/operators.js:244
  res.meta[key] = value
                ^

TypeError: Cannot set property 'live' of undefined
    at updateMeta (/home/arj/dev/ssb-db2/node_modules/jitdb/operators.js:244:17)
    at /home/arj/dev/ssb-db2/node_modules/jitdb/operators.js:298:41
    at /home/arj/dev/ssb-db2/node_modules/jitdb/operators.js:277:48
    at Array.map (<anonymous>)
    at or (/home/arj/dev/ssb-db2/node_modules/jitdb/operators.js:277:6)
    at Object.<anonymous> (/home/arj/dev/ssb-db2/test.js:18:5)
```

The problem comes from [here](https://github.com/ssb-ngi-pointer/jitdb/blob/811cba90652ff13052fac0613550b33f58423784/operators.js#L255) because we run the and (waiting, live) before fromDB has had a chance to set meta. I was thinking we could fix it by changing query in db.js from:

```
      return jitdbOperators.query(ops, and(waitUntilReady, first), ...rest)
```

to

```
      return jitdbOperators.query(ops, and(waitUntilReady), first, ...rest)
```

this then breaks the test for `or(hasRoot('%something'))`. So I'm not entirely sure what the best way to fix this case is.

PR #135, opened 2021-01-21
support db.query(live(), toPS())

staltz:
Final bits for #133

### DIFF
--- a/test/operators.js
+++ b/test/operators.js
@@ -473,23 +473,40 @@ test('live votesFor', (t) => {
         db.query(
           and(votesFor(postMsg.key)),
           live({ old: true }),
-          toPullStream(),
-          pull.drain(
-            (result) => {
-              if (i++ == 0) {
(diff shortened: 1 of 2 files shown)


arj03:
Oh yes, easy merge.

test/operators.js
```diff
@@ -473,23 +473,40 @@ test('live votesFor', (t) => {
         db.query(
           and(votesFor(postMsg.key)),
           live({ old: true }),
```
staltz, line 475: @arj03 Check below. This was actually using the API wrong, but it happened to function correctly!

Before, it was 

```js
pull( 
  db.query(stuff, toPullStream(), pull.drain(etc)
)
```

When it should have been:


```js
pull( 
  db.query(stuff, toPullStream()), 
  pull.drain(etc)
)
```

arj03: Yeah I noticed :-)

PR #136, opened 2021-01-21
Add script with a bunch of common queries

arj03:
I noticed that I didn't have a very good idea of how common queries are running. This gives us a better idea. Might be good to run in the CI thingy. But for now I'll just dump them here.

Output

```
running: key initial
query: 199.9ms
running: key 2
query: 25.106ms
running: key again
query: 22.5ms
running: latest root posts
query: 125.503ms
running: latest posts
query: 233.988ms
running: votes initial
query: 60.99ms
running: votes 2
query: 16.416ms
running: votes again
query: 7.335ms
running: author posts
query: 309.867ms
running: author posts again
query: 65.469ms
```


staltz:
Could we merge this PR only when it's running against ssb-fixtures? This way we can objectively compare numbers across computers, and we don't need to depend on personal data (and their msg.keys), also making it better for other future contributors to run them.

arj03:
Yes very good idea. Was mostly interested in getting in some good queries that are used a lot in apps.

staltz:
Or actually, how about folding these into the benchmark/index.js? After all, the goal of these scripts is to check the perf. Then it's easier, just add new tests with these queries

arj03:
Agree, that's what I meant with the: "Might be good to run in the CI thingy" :-) I'm using this now mostly to test the optimization for prefixes. Once that is done I'll make a proper PR. This was mostly meant as an open invitation if you have any queries from your threads testing but you can of course always add them afterwards :)

benchmark/index.js
```diff
@@ -156,3 +169,84 @@ test('initial indexing', async (t) => {
...
+    startFrom(0),
+    paginate(25),
+    descending(),
+  ],
+
+  'votes one initial': [and(votesFor(KEY1))],
+
+  'votes one': [and(votesFor(KEY1))],
+
+  'votes again': [and(votesFor(KEY1))],
```
staltz, line 203: @arj03 see above, can we delete one of these? Running it three times doesn't seem necessary

staltz:
@arj03 I pushed some commits that move these queries into benchmark/index.js

benchmark/index.js
arj03: Yes

arj03:
The idea with the second run is that we test not having to load the index from disc. So it needs to be a another key otherwise we will test your weakmap cache instead.

arj03:
Other than that it's really good to see these queries here in the CI

staltz (approved):

staltz:
Alright, LGTM. How about you review and merge? (note that perf.js was deleted)

PR #137, opened 2021-01-25
some operators use useMap and prefix

staltz:
Before and after of the perf benchmarks:

```diff
  migration (using ssb-db)

    ✔ delete db2 folder to start clean
    ✔ ssb-db has finished indexing
-   ✔ duration: 2816ms
+   ✔ duration: 2870ms
    fallback to close

  migration (alone)

    ✔ delete db2 folder to start clean
-   ✔ duration: 2705ms
+   ✔ duration: 2722ms

  initial indexing

    ✔ null
-   ✔ duration: 2008ms
+   ✔ duration: 2043ms

  setup

  key one initial
-   ✔ duration: 454ms
+   ✔ duration: 464ms

  key two
-   ✔ duration: 19ms
+   ✔ duration: 20ms

  key one again
-   ✔ duration: 14ms
+   ✔ duration: 15ms

  latest root posts
-   ✔ duration: 371ms
+   ✔ duration: 364ms

  latest posts
-   ✔ duration: 29ms
+   ✔ duration: 32ms

  votes one initial
-   ✔ duration: 355ms
+   ✔ duration: 383ms

  votes again
-   ✔ duration: 11ms
+   ✔ duration: 0ms

  hasRoot
-   ✔ duration: 302ms
+   ✔ duration: 294ms

  hasRoot again
-   ✔ duration: 12ms
+   ✔ duration: 0ms

  author one posts
-   ✔ duration: 248ms
+   ✔ duration: 249ms

  author two posts
-   ✔ duration: 61ms
+   ✔ duration: 66ms

  teardown

  total:     18
  passing:   18
  duration:  15.3s
```

### DIFF
--- a/operators/index.js
+++ b/operators/index.js
@@ -47,6 +47,7 @@ function votesFor(msgKey) {
     type('vote'),
     equal(seekVoteLink, msgKey, {
       prefix: 32,
+      useMap: true,
       indexType: 'value_content_vote_link',
     })
   )()
@@ -57,6 +58,7 @@ function contact(feedId) {
     type('contact'),
     equal(seekContact, feedId, {
       prefix: 32,
+      useMap: true,
       indexType: 'value_content_contact',
     })
   )()
@@ -72,6 +74,7 @@ function mentions(key) {
 function hasRoot(msgKey) {
   return equal(seekRoot, msgKey, {
     prefix: 32,
+    useMap: true,
     indexType: 'value_content_root',
   })
 }
@@ -79,6 +82,7 @@ function hasRoot(msgKey) {
 function hasFork(msgKey) {
   return equal(seekFork, msgKey, {
     prefix: 32,
+    useMap: true,
     indexType: 'value_content_fork',
   })
 }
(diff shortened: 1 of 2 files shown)


arj03:
Sweet. I had a similar PR here locally. Wanted to get the concurrent update in, but this is good. We should use prefix for key as well. It's about 2x faster. But there is still quite a bit of key collision (around 1/5) so I'll try and maybe do the % trick and see if that helps. Anyway this needs a benchmark update as well, so lets just get this in first and I'll look at key afterwards.

staltz:
> Sweet. I had a similar PR here locally. Wanted to get the concurrent update in, but this is good.

Oh, I didn't know you had a branch. Sorry about that.

> We should use prefix for key as well. It's about 2x faster.

`key` operator currently uses prefix indexes (but not `useMap`), so what makes it 2x faster?

arj03:
Don't worry about the branch, it was mostly for testing this stuff.

Key is 2x faster with useMap

staltz:
Interesting! I thought `key()` would be like `author()`, and that `author()` doesn't benefit from useMap, see your comment here: https://github.com/ssb-ngi-pointer/jitdb/pull/95#issuecomment-765467232

arj03:
But maybe we should add that as an option? prefixOffset?

staltz:
> But maybe we should add that as an option? prefixOffset?

I was typing this right now. :)
I was going to suggest `prefixStart` but `prefixOffset` is better.

arj03:
Cool, will work on that after lunch.

PR #143, opened 2021-01-26
add benchmark for two indexes updating concurrently

staltz:
@arj03  What do you think? Is this bench useful? Not necessary to merge.


arj03:
Sure why not :-)

PR #144, opened 2021-01-27
emit event ssb:db2:indexing:progress

staltz:
Similar to `ssb:db2:migrate:progress`. I thought that this number crunching would be likely needed in many apps, so better to put it here. If people want more precise numbers, per index, they can use `getStatus`.

### DIFF
--- a/db.js
+++ b/db.js
@@ -86,7 +86,7 @@ exports.init = function (sbot, config) {
   // Crunch stats numbers to produce one number for the "indexing" progress
   status.obv((stats) => {
     const logSize = Math.max(1, stats.log) // 1 prevents division by zero
-    const nums = Object.values(stats.jit).concat(Object.values(stats.indexes))
+    const nums = Object.values(stats.indexes)
     const N = Math.max(1, nums.length) // 1 prevents division by zero
     const progress = Math.min(
       nums
--- a/db.js
+++ b/db.js
@@ -83,6 +83,21 @@ exports.init = function (sbot, config) {
     stateFeedsReady.resolve()
   })
 
+  // Crunch stats numbers to produce one number for the "indexing" progress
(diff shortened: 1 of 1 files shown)


arj03:
Lets see the numbers for these

arj03:
Seems like we got a fast benchmark machine this time :)

arj03:
This looks good. One question, what happens to jitdb indexes that are not updated? Will they keep the indexing from reaching 100%?

staltz:
> One question, what happens to jitdb indexes that are not updated?

When could this happen? So far I haven't seen (e.g. in `npm run benchmark-no-create`) a case where it doesn't reach 100%

arj03:
I was thinking that you had say a isPrivate index and you where on the main screen. Then unless you go to the private view and actually do the query, the index will not be updated.

staltz:
Ah, right, but then on the main screen that index would be non-existent and not taken into account in the calculations. Only when it exists in `stats.jit` then it would be taken into calculations. Or can you foresee the possibility that an index remains "partially indexed"? e.g. `stats = {log: 1000, {jit: {private: 450}}`, (private remains stuck in 450)

staltz:
Oh wait a minute, now I think I know what you're talking about. Hmmm

Any suggestions?

arj03:
I would leave out the jitdb indexes in a progress.

staltz:
Yeah, I can do that, the eager indexes are the leveldb ones, but we'd still could/should have some kind of progress indicators for JITDB, because they can take actually a good while to fully index. Easily at least 10 seconds for all the queries done on the public tab for all the posts and votes, and hasRoot, isPublic, etc. Maybe not in this PR, but we need something.

arj03 (approved):

PR #148, opened 2021-01-27
add test for hookable createHistoryStream

staltz:
1st adds failing test, 2nd commit fix.

We need this because ssb-replicate/legacy does a hook on the history stream.


arj03:
Ouch, the dreaded triangle.

arj03 (approved):

PR #150, opened 2021-01-28
report full memory usage in benchmarks

staltz:
`.rss` covers everything, while `.heapTotal` is just objects, arrays, strings, buffers, things like that. I believe we have a LOT of data put in the stack, because we heavily rely on small numbers and small strings.

I checked in `htop`, both desktop and mobile (yeah you can do `htop` in Termux on Android) and memory usage of Node.js on desktop and of Manyverse on Android (using the latest ssb-db2) float around half a gigabyte.

### DIFF
--- a/benchmark/index.js
+++ b/benchmark/index.js
@@ -63,7 +63,7 @@ if (!skipCreate) {
 
 let maxRAM = 0
 function updateMaxRAM() {
-  maxRAM = Math.max(maxRAM, process.memoryUsage().heapTotal)
+  maxRAM = Math.max(maxRAM, process.memoryUsage().rss)
 }
 
 test('migration (using ssb-db)', async (t) => {
--- a/benchmark/index.js
+++ b/benchmark/index.js
@@ -61,9 +61,22 @@ if (!skipCreate) {
   })
 }
 
-let maxRAM = 0
+let maxRAMrss = 0
+let maxRAMheap = 0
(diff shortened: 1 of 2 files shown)


arj03:
Hmm that is a lot more.

staltz:
I suspected this would happen (high mem usage) because we're heavily relying on in-memory maps and caches and stuff like that, to speed up processing time, and in contrast to Leveldb, we load everything up, not in chunks.

arj03:
Yeah it would be good to identity like how much of this is jitdb for example.

arj03:
Maybe we can report both numbers in the benchmark?

staltz:
Updated PR. Now it should report 4 things:

- `rss` and `heapUsed` when only `db.js` and `migrate.js` are used (during migrate tests)
- maximum `rss` / `heapUsed`

Note that `heapUsed` is different from the previous `heapTotal` we were using. I'm still a bit puzzled about these different pieces of mem usage, but I think `heapTotal` are those things in JS that **we control** *after* garbage collection has remove garbage.

arj03 (approved):

arj03:
Interesting. Especially that memory usage without indexes. That gives us some hope that we can get that down without changing too much. I'd like to dig more into this and see what I can find. Lets merge this for now. Really starting to appreciate the CI reporting here.

PR #155, opened 2021-01-29
hack to fix issue 154

staltz:
It's ugly but it works :disappointed:


arj03:
Maybe a better fix would be an option to EBT so it doesn't overwrite post, because if you are still using db1 to write messages into, then better use that. Maybe even just don't overwrite post if it's already there?

staltz:
Yeah, sounds good, that didn't cross my mind.

PR #157, opened 2021-02-01
refactor: avoid pyramid of ifs and ~ operator

staltz:
Just a cosmetic thing. 

I stumbled upon this while I was thinking about making another level index for abouts such as `{type:'put', key: [content.about, "name"], value: processed}` (might make a PR for that this week, I think it'll be much more efficient than looking for **all** about msgs, what ssb-social-index does)

### DIFF
--- a/indexes/full-mentions.js
+++ b/indexes/full-mentions.js
@@ -35,37 +35,31 @@ module.exports = function (log, dir) {
     const recBuffer = record.value
     if (!recBuffer) return batch.length // deleted
 
+    const pKey = bipf.seekKey(recBuffer, 0, bKey)
+
     let p = 0 // note you pass in p!
-    const pKey = bipf.seekKey(recBuffer, p, bKey)
-
-    p = 0
     p = bipf.seekKey(recBuffer, p, bValue)
-    if (~p) {
-      const pContent = bipf.seekKey(recBuffer, p, bContent)
-      if (~pContent) {
-        const pMentions = bipf.seekKey(recBuffer, pContent, bMentions)
-        if (~pMentions) {
-          const mentionsData = bipf.decode(recBuffer, pMentions)
-          if (Array.isArray(mentionsData)) {
-            const shortKey = bipf.decode(recBuffer, pKey).slice(1, 10)
-            mentionsData.forEach((mention) => {
-              if (
-                mention.link &&
-                typeof mention.link === 'string' &&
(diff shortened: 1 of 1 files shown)


arj03 (approved):

arj03:
social index uses the dest to pick out the right abouts. The browser core actually had a simple level index of author -> latest value for self. Interesting enough loading just this index on startup with ~1000 profiles turned out to take roughly 4 seconds. I do think the heaviest part in all of this is the description, which is why in the latest social index version I only load that when needed (when looking at their profile).

arj03:
One of the reason you might be experiences worse performance is that ssb-suggest uses friends (hops=1) and in the browser I don't have friends (yet, working on it), besides I don't follow that many accounts there. That said it should not be worse than db1.

Another thing is that ssb-suggets does [names](https://github.com/ssbc/ssb-suggest/blob/1d307ece57e84ee3643b684bb4b6463861fd5784/lib/avatar.js#L19) and at least I have no use for that. So I'm thinking of adding a flag to disable that. Do you use that in manyverse?

Another thing that is a bit silly is that there is a cache in ssb-suggest (suggestCache) and then at least I also have a cache for when using about directly. It would be much better with only a single cache.

staltz:
>  I do think the heaviest part in all of this is the description, which is why in the latest social index version I only load that when needed (when looking at their profile).

Interesting, makes sense! 

> had a simple level index of author -> latest value for self

Yeah I was going to do this, but I'm not sure if it's going to be faster than JIT.

> Another thing is that ssb-suggets does names and at least I have no use for that. So I'm thinking of adding a flag to disable that. Do you use that in manyverse?

I don't use that. In fact, I'd much prefer if we just use the person's own self-assigned abouts than the "social value" abouts. It's simpler, faster, and also as a community we haven't had much benefit from social value abouts.

> also have a cache for when using about directly

In Manyverse I use [ssb-cached-about](https://www.npmjs.com/package/ssb-cached-about). I also use ssb-about directly, so I've tested performance in both ways.

arj03:
The thing I found when I was doing the self-assigned name thing was that not everyone has a name then. It seems like some people used they main profile to assign name and pictures to they mobile client.

staltz:
Yeah, I believe that can happen. I'm curious to try only self-assigned names, though, because there will also be a lot of variation in SSB with the patterns of how msgs are authored, hard to "catch all patterns" and it comes with a performance cost too.

PR #159, opened 2021-02-02
add about() operator

staltz:
I'm going to use this in a lite alternative of ssb-about, but also this stuff could be used directly inside ssb-social-index.


package.json
```diff
@@ -44,6 +44,7 @@
     "rimraf": "^3.0.2",
     "secret-stack": "6.3.1",
     "ssb-caps": "1.1.0",
+    "ssb-config": "^3.4.5",
```
arj03, line 47: this doesn't look like its used?

staltz: Oh, sorry, I used that for a non-committed example.js where I'm testing on my production ssb. I can remove it if you wish

arj03: Yeah lets remove that :)

arj03 (approved):

PR #160, opened 2021-02-02
Fix add/publish so we wait for the state to be loaded

arj03:
This fixed the problem found in https://github.com/ssbc/ssb-friends/pull/49. Namely that we don't wait for the state to be loaded, so we end up with the wrong validate. Only publish and add really need this anyway which is why they are the only ones with this check.

### DIFF
--- a/test/basic.js
+++ b/test/basic.js
@@ -17,14 +17,14 @@ mkdirp.sync(dir)
 
 const keys = ssbKeys.loadOrCreateSync(path.join(dir, 'secret'))
 
-const sbot = SecretStack({ appKey: caps.shs })
+let sbot = SecretStack({ appKey: caps.shs })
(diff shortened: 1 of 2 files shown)


arj03:
Must have been the random failure we have seen one time before. Works locally.

test/basic.js
```diff
@@ -157,11 +157,57 @@ test('delete all', (t) => {
             (err, results) => {
               t.error(err, 'no err')
               t.equal(results.length, 0, 'gone')
-              sbot.close(t.end)
+              t.end()
```
staltz, line 160: Since we're creating a new sbot, couldn't we close this previous one?

staltz: Nevermind!

staltz (approved):
Nice find!

PR #164, opened 2021-02-08
add benchmarks for all migrate situations

staltz:
Adds new API:

- `sbot.db2migrate.stop` (needed for one new benchmark)

Adds new benchmark tests:

- Run migrate together with db1 and db2
- Run migrate and db2, but no db1
- Run migrate and db2, continuing from a 90%-completed previous migration, to test migrate's `findMigratedOffset` function

Optimizes migrate's `inefficientFindMigratedOffset`:

- This was doing a log.stream, but it never read the contents of the records, so this PR adds a `decrypt: false` opt that bypasses private.decrypt
  - Even though private.decrypt has some checks that skip the majority of unbox calls, this new optimization skips the unbox call also when the msg is intended for the author (i.e. canDecrypt)
  - This improves the "Migrate continuation (+db2)" benchmark run by -15%

### DIFF
--- a/benchmark/index.js
+++ b/benchmark/index.js
@@ -245,12 +245,14 @@ test('migrate continuation (+db2)', async (t) => {
 
   pull(
     fromEvent('ssb:db2:migrate:progress', sbot),
-    pull.filter((progress) => progress > 0.5),
+    pull.filter((progress) => progress > 0.9),
     pull.take(1),
     pull.drain(async () => {
+      sbot.db2migrate.stop()
       await sleep(4000) // wait for new log FS writes to finalize
       await new Promise((r) => sbot.close(() => r()))
       await sleep(500) // some silence
+      t.pass('migrated 90%, will reset sbot')
 
       sbot = SecretStack({ appKey: caps.shs })
         .use(require('../'))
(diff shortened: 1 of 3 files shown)


benchmark/index.js
```diff
@@ -165,6 +162,128 @@ test('migration (alone)', async (t) => {
...
+
+  pull(
+    fromEvent('ssb:db2:migrate:progress', sbot),
+    pull.filter((progress) => progress === 1),
+    pull.take(1),
+    pull.drain(async () => {
+      const duration = Date.now() - start
+      t.pass(`duration: ${duration}ms`)
+      fs.appendFileSync(reportPath, `| Migrate (+db1 +db2) | ${duration}ms |\n`)
+      await sleep(4000) // wait for new log FS writes to finalize
```
arj03, line 189: can you do a sbot.db.onDrain here?

arj03:
The review note is just a small thing. Might make it less prone to random failures.

Great to see the optimization on migrate continue, the option could be useful in other settings.

arj03 (approved):

benchmark/index.js
staltz: Oh yeah, good idea, I was copied this `await sleep(4000)` from previous tests above, but those didn't have db2 for me to call sbot.db.onDrain. I'll do that here

PR #165, opened 2021-02-08
add sbot.ready() to compat/db

staltz:
ssb-replicate uses this


arj03:
Was trying to find how this is set in ssb-db but got lost in the code base :-) From how it is used in ssb-replicate this PR looks fine :)

arj03 (approved):

staltz:
Yeah, I thought it's safe to always return true because in db2 we do an implicit onDrain inside the queries, so we don't need the developer to manually check for a ready boolean.

PR #168, opened 2021-02-10
tweak too-hot params

staltz:
See https://github.com/ssb-ngi-pointer/jitdb/issues/118#issuecomment-777096352

This was a clear win, I think.

### DIFF
--- a/log.js
+++ b/log.js
@@ -51,7 +51,7 @@ module.exports = function (dir, config, privateIndex) {
   log.stream = function (opts) {
     const shouldDecrypt = opts.decrypt === false ? false : true
     const tooHot = config.db2.maxCpu
-      ? TooHot({ ceiling: config.db2.maxCpu, maxPause: 1000 })
+      ? TooHot({ ceiling: config.db2.maxCpu, wait: 90, maxPause: 300 })
       : () => false
     const s = originalStream(opts)
     const originalPipe = s.pipe.bind(s)


arj03:
Sweet! I'm wondering if we should make a benchmark here that uses maxcpu so we can measure this?

arj03 (approved):

arj03:
Still the same as in the other PR?

staltz:
Locally I get

**Before**

```
  initial indexing

    ✔ null
    ✔ duration: 886ms

  initial indexing maxcpu 86

    ✔ delete indexes folder to start clean
    ✔ null
    ✔ duration: 5042ms
```

**After**

```
  initial indexing

    ✔ null
    ✔ duration: 930ms

  initial indexing maxcpu 86

    ✔ delete indexes folder to start clean
    ✔ null
    ✔ duration: 4817ms
```

You're right, there's not a big enough difference. Maybe there are two things going on: (1) with 1million records, it's makes a bigger diff, (2) muxrpc weirdness again? (no muxrpc in these benchmarks)

arj03:
Locally on my machine there is almost no difference. It seems to be almost slower with these changes. I'm getting 24s without this and 25s afterwards. This is creating the author index on a full database (1.5million)

arj03:
And still as I said in the other comment, I'm learning towards creating these indexes with no max cpu, but then again I don't really use this option.

staltz:
It's weird, I was sure this did a change yesterday when I ran it locally...

staltz:
I think this clearly affects Manyverse desktop (also mobile but I'm running the numbers on desktop), for instance, I deleted `value_author.32prefix` before running the app, I ran it 3 times:

**Before**

```
jitdb missing indexes: [ { type: 'EQUAL', data: { seek: [Function: seekAuthor] ...
jitdb growing index +185ms
itdb growing index +44ms
jitdb growing index +240ms
jitdb growing index +425ms
jitdb growing index +5s
jitdb growing index +12s
jitdb growing index +26s
jitdb log.stream #741 done 1255213 records in 55597ms +12s
```

```
jitdb log.stream #429 done 1255213 records in 64716ms +13s
```

```
jitdb log.stream #893 done 1255213 records in 55267ms +10s
```

**After** setting `wait=90`, `maxPause=300`

```
jitdb log.stream #436 done 1255213 records in 13369ms +2s
```

```
jitdb log.stream #510 done 1255213 records in 9482ms +2s
```

```
jitdb log.stream #979 done 1255213 records in 10026ms +2s
```

arj03:
Really wierd. I don't really understand why which is probably was is bugging me the most. But very happy that it is such a big improvement :)

staltz:
So is it fine to merge? I do agree we should get to the bottom of it at some point

arj03:
Yes please do

PR #169, opened 2021-02-11
benchmark indexing with maxCpu=86

staltz:
This should be merged before https://github.com/ssb-ngi-pointer/ssb-db2/pull/168, and then we can rebase that one and see the new benchmark results


arj03:
Nice!

PR #171, opened 2021-02-12
Add self-assigned about index

arj03:
Adds an index for self-assigned about messages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,28 @@ sbot.db.query(
 )
 ```
 
+Another extra index plugin that is commonly needed in SSB communities
+is the **about-self** index. This indexes only self-assigned about
+messages in contrast to [ssb-social-index] that indexes all about
+messages.
+
+Example usage:
+
+```js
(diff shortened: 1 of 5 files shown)


staltz:
What was the impact that `description` had on performance (runtime but also storage)?

arj03:
It was actually surprisingly little. Both in terms of indexing performance but also on load. If I remember correctly it took something like 120ms to load this on boot without description and 150ms with description.

arj03:
These tests failures are starting to become problematic. Just restarted it 3-4 times...

staltz:
I recommend we do a `test.skip` for that `config.maxCpu makes indexing last longer` test that is flaky.

staltz:
PR looks good, but do you mind if I work on this branch to add some operator for this? I'd like to have both "get" and live stream APIs for self abouts.

arj03:
Sure, go ahead :)

staltz:
Actually I might do it in a separate PR. So merging :)

PR #172, opened 2021-02-12
Add failed EBT experiment

arj03:
Yesterday I tried a thing. I noticed that EBT is one of the slowest indexes, and looking at it, it's pretty clear why, it basically has to store everything message in level. On my machine EBT is 12 seconds to index from scratch. I was wondering if I could use some of the same things as I used for ssb-friends on this index and tried that. This takes the full index time down to 7 seconds, but it adds a startup overhead of 700ms, so I don't think its worth it. It does reduce the size of the database from 24mb to 1.4mb. 

So this PR is mainly a learning thing and can be closed.

### DIFF
--- a/indexes/ebt.js
+++ b/indexes/ebt.js
@@ -1,5 +1,8 @@
 const bipf = require('bipf')
+const pull = require('pull-stream')
+const pl = require('pull-level')
 const Plugin = require('./plugin')
+const jsonCodec = require('flumecodec/json')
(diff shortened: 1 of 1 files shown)


staltz:
Ok, before closing, let me take a look at it to understand, and maybe we find another insight.

staltz:
How about we mark this PR as a draft to signal that it's not supposed to be merged? (you can undo the draft state if you disagree)

PR #173, opened 2021-02-12
Fix processed

### LINKED ISSUE #167: Level indexes have different `processed`
This shows in Manyverse (Android) when starting up the app. Shouldn't all these `processed` seqs be the same?

```
ssb:db2:base got index status: { version: 1, offset: 767860823, processed: 2365509 }
ssb:db2:ebt got index status: { version: 1, offset: 767860823, processed: 1349552 }
ssb:db2:contacts got index status: { version: 2, offset: 767860823, processed: 1016108 }
ssb:db2:fullMentions got index status: { version: 1, offset: 767860823, processed: 1124452 }
```

arj03:
Fix #167. It turns out that full mentions was using this processed. So it was buggy before. I added a test to show that it was buggy before, but is now working. Also makes the indexes not have to do the "can I use this" check.

### DIFF
--- a/indexes/base.js
+++ b/indexes/base.js
@@ -40,7 +40,6 @@ module.exports = function (log, dir, privateIndex) {
   }
 
   function handleData(record, processed) {
-    if (record.offset < offset.value) return batch.length
     const buf = record.value
     if (!buf) return batch.length // deleted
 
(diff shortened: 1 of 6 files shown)


arj03:
Bahh, something broke the tests. Will have to look at that. Sorry for the noise.

arj03:
Now this is back to the maxcpu test. Should be ready for review.

staltz (approved):

arj03:
Thanks for the review

PR #175, opened 2021-02-12
Refactor plugin pattern as OOP

staltz:
This is a bit debatable PR, because it's a big refactor the plugins. But I'd like to explain the purpose: basically to simplify plugin creation, not just for us, but specially for other developers in the future. I didn't bring this up in our call because I suddenly thought about trying this. :)

Plugin developers had to be conscious about several details: must remember to put `close: level.close.bind(level)` at the bottom, must re-export fields like `onData` and `writeBatch` even if the new plugin doesn't call these things, and overall there's some duplicate code going on.

I thought about what to do here (not just today, I've been thinking about this randomly in the past weeks), and I think this is where OOP and inheritance helps. I don't mind the closure and `this`less pattern, I'm fine with that. But there are use cases where it's not handy enough, where OOP is. (I also know of cases where OOP is not handy and the closure pattern is!)

So the basic idea here is that plugin developers write this:

```js
module.exports = class MyNewIndex extends Plugin {
  constructor(log, dir) {
    // log, dir, name, version, keyEncoding, valueEncoding
    super(log, dir, 'myindex', 1, 'json', 'json')
  }

  processRecord(record, seq) {
    // implement me
    // and then do `this.batch.push({...})`
  }

  myVeryOwnMethodToGetMyData(x, cb) {
    // implement me
  }
}
```

Nothing else is needed, no `writeData`, no re-exports. The constructor takes care of basic configurations of leveldb, and then there's always a `processRecord` method (used to be called `handleData`) and then whatever methods the plugin developer wants. Note that this PR is a net negative in lines of code (-110 LOC), shows how some duplication got removed.

I also did a bunch of renames to make things consistent with each other. E.g. `data` is too vague, I used `record` when it refers specifically to the log's buffer and seq and offset. I renamed `handleData` to `processRecord` to be consistent with the variable `processed`. And some other stuff.

Let me know what you think of this, I know it's a bit too much code to do a review, and I'm not in a hurry for a merge.

### DIFF
--- a/indexes/about-self.js
+++ b/indexes/about-self.js
@@ -12,7 +12,7 @@ const bAbout = Buffer.from('about')
 // feedId => hydratedAboutObj
 module.exports = class AboutSelf extends Plugin {
   constructor(log, dir) {
-    super(dir, 'aboutSelf', 3, 'json', 'json')
+    super(log, dir, 'aboutSelf', 3, 'json', 'json')
     this.profiles = {}
   }
 
--- a/indexes/about-self.js
+++ b/indexes/about-self.js
@@ -1,72 +1,63 @@
 const bipf = require('bipf')
 const pull = require('pull-stream')
 const pl = require('pull-level')
-const jsonCodec = require('flumecodec/json')
-const Plugin = require('ssb-db2/indexes/plugin')
+const Plugin = require('./plugin')
 
-// 1 index:
-// - feed => hydrated about
+const bValue = Buffer.from('value')
+const bAuthor = Buffer.from('author')
+const bContent = Buffer.from('content')
+const bType = Buffer.from('type')
+const bAbout = Buffer.from('about')
 
-module.exports = function (log, dir) {
-  const bValue = Buffer.from('value')
-  const bAuthor = Buffer.from('author')
-  const bContent = Buffer.from('content')
-  const bType = Buffer.from('type')
-  const bAbout = Buffer.from('about')
-
-  let batch = []
-
-  const name = 'aboutSelf'
-  const { level, offset, stateLoaded, onData, writeBatch } = Plugin(
-    dir,
-    name,
-    3,
-    handleData,
-    writeData,
-    beforeIndexUpdate
-  )
-
-  function writeData(cb) {
-    level.batch(batch, { keyEncoding: 'json', valueEncoding: 'json' }, cb)
-    batch = []
+// feedId => hydratedAboutObj
+module.exports = class AboutSelf extends Plugin {
+  constructor(log, dir) {
+    super(dir, 'aboutSelf', 3, 'json', 'json')
+    this.profiles = {}
   }
 
-  function handleData(record, processed) {
-    if (record.offset < offset.value) return batch.length
-    const recBuffer = record.value
-    if (!recBuffer) return batch.length // deleted
+  onLoadedMeta(cb) {
+    pull(
+      pl.read(this.level, {
+        gte: '',
+        lte: undefined,
+        keyEncoding: this.keyEncoding,
+        valueEncoding: this.valueEncoding,
+        keys: true,
+      }),
+      pull.drain((data) => (this.profiles[data.key] = data.value), cb)
+    )
+  }
+
+  processRecord(record, seq) {
+    const buf = record.value
 
     let p = 0 // note you pass in p!
-    p = bipf.seekKey(recBuffer, p, bValue)
-    if (!~p) return batch.length
+    p = bipf.seekKey(buf, p, bValue)
+    if (p < 0) return
+    const pAuthor = bipf.seekKey(buf, p, bAuthor)
+    const pContent = bipf.seekKey(buf, p, bContent)
+    if (pContent < 0) return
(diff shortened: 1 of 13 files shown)


staltz:
Benchmark fails on the `ssb-threads and ssb-friends` test because we have a chicken and egg situation: we first have to update ssb-friends to use the new Plugin class before we can run benchmarks on it.

arj03:
Not against this, but need to some to look through the changes properly :)

staltz:
Additional information (this will eventually go into README):

Subclasses can customize some behavior by implementing methods that may be implemented in the superclass, e.g.

- `onLoaded` (this used to be `beforeIndexUpdate`), see an example in `about-self`
- `flushBatch` (this used to be `writeData`), see an example in `base`

indexes/about-self.js
```diff
@@ -1,72 +1,63 @@
 const bipf = require('bipf')
 const pull = require('pull-stream')
 const pl = require('pull-level')
-const jsonCodec = require('flumecodec/json')
```
arj03, line 4: Careful with this. I'm quite sure this one still needs to be there. Pull-level is really wierd with 'json' and jsonCodec, I remember spending at least an hour tracking down a bug related to that.

arj03:
Overall I like this. Anything that removes duplicated code (as seen by the net positive removals) is a good sign. I added one comment that needs to be fixed first. And as you explained a README update. Other than that should be good to go.

indexes/about-self.js
staltz: Sure. What happens is that level-codec accepts either a string or an object. So either `json` or `require('flumecodec/json')`.

https://github.com/Level/codec/blob/fdb2c4fb2db2fe0f51be7d221fd74ee70794a794/index.js#L13-L17

```js
  if (typeof encoding === 'string') encoding = encodings[encoding]
  if (!encoding) encoding = encodings.id
  return encoding
```

The "encodings" above is this: https://github.com/Level/codec/blob/v9.0.2/lib/encodings.js

`flumecodec` is just a thin 1-LOC wrapper around `level-codec`: https://github.com/flumedb/flumecodec/blob/master/json.js IMO we should forget `flumecodec`, it's totally useless.

Note that tests in this PR were not changed, and that we had both `jsonCodec` (imported object) and `'json'` (string) mixed sometimes, and it worked, see: https://github.com/ssb-ngi-pointer/ssb-db2/blob/f7639e9225458f4a342e379a5f4e2384150056db/indexes/about-self.js#L30 and https://github.com/ssb-ngi-pointer/ssb-db2/blob/f7639e9225458f4a342e379a5f4e2384150056db/indexes/about-self.js#L89

So, moving forwards, let's just confidently use `keyEncoding: 'json'` or `keyEncoding: 'utf8'` or `keyEncoding: oneOfThoseStrings` and forget `flumecodec`.

arj03: Hmm, but there is no test of this close and load again in aboutSelf. Maybe we should add that then?

staltz: `level-post` does not support string encoding: https://github.com/dominictarr/level-post/blob/9021f153f159aa1d13233908ebe616f650e78f22/index.js#L17

staltz:
@arj03 I expected https://github.com/ssb-ngi-pointer/ssb-db2/pull/175/commits/860b0545b3cd5f7b2b0a91ecb4571bece4d0b235 to fail the newly added test, but... it didn't?

arj03:
Odd. I can have a look a bit later today.

staltz:
Ok, the test does not trigger the bug because the bug only occurs when we use `pull-level` in live mode **and** when doing a `put` or `del`. In fact, `level-post` only uses `keyEncoding.encode`, not `keyEncoding.decode`.

I already have an idea how to fix this in the implementation, but I'd still like to get a test case that triggers the bug.

staltz:
I found a good test case: `full-mentions` index in `live` mode. Currently there's no tests for that. However, it's not as simple as adding a test, because it turns out `full-mentions` is full of bugs. One bug: it seems to always do "live and old" even if you pass `meta.live = 'liveOnly`.

staltz:
I'll put this PR on hold and work on fixing full-mentions.

staltz:
Rebased on master. It should fail the test now. (commit to fix it coming soon)

staltz:
Alright, this is ready. Note these commits:

- https://github.com/ssb-ngi-pointer/ssb-db2/pull/175/commits/2599ab85736596f549c41b2311ac3989a7c4161e fixes the codec weirdness
- https://github.com/ssb-ngi-pointer/ssb-db2/pull/175/commits/aada807abecc10f3b5ede9e0a54e4c9502247fab some minor refactor
- https://github.com/ssb-ngi-pointer/ssb-db2/pull/175/commits/eea63d3c12b011da6f80604cd0b092300c23ccf5 readme

arj03:
This looks good. I checked the new commits. Thanks for cleaning this up. Feels good to have the README update with plugins as well. Should come in handy in the future.

arj03 (approved):

staltz:
Thanks for reviewing this big one!

PR #177, opened 2021-02-15
add test for full-mentions in live mode

staltz:
1st test, 2nd fix


staltz:
Note, the solution is just to add `()`. This was certainly a footgun, I'm sorry that I made that in JITDB. I have an idea how to improve JITDB to make this case not happen again (i.e. it should work with or without `()`)` and it's a small PR, I'll send it soon.

arj03:
Oh damn, nice catch

arj03 (approved):

PR #178, opened 2021-02-15
createWriteStream

staltz:
1st test, 2nd fix

### DIFF
--- a/test/createHistoryStream.js
+++ b/test/createHistoryStream.js
@@ -17,6 +17,7 @@ const keys = ssbKeys.loadOrCreateSync(path.join(dir, 'secret'))
 
 const sbot = SecretStack({ appKey: caps.shs })
   .use(require('../'))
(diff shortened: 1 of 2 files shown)


test/createHistoryStream.js
```diff
@@ -69,6 +70,42 @@ test('No values', (t) => {
   )
 })
 
+test('createWriteStream', (t) => {
+  const rando = ssbKeys.generate()
```
arj03, line 74: :-)

staltz: This should fix the "rando problem"

arj03 (approved):

arj03:
Sweet. Nice and simple.

staltz:
I might later make a PR to use `pull-drain-gently` inside createWriteStream (obviously this would be guarded by a check whether `config.maxCpu` exists), but if and only if I discover it helps manyverse during initial replication.

PR #179, opened 2021-02-15
rework keys index

staltz:
So far this just adds a new benchmark. Will add implementation code later.

### DIFF
--- a/benchmark/index.js
+++ b/benchmark/index.js
@@ -81,11 +81,16 @@ function reportMem() {
   return `${rss} MB = ${heap} MB + etc`
 }
 
+let keys
+test('setup', (t) => {
+  keys = ssbKeys.loadOrCreateSync(path.join(dir, 'secret'))
+  t.end()
+})
+
 test('migrate (+db1)', async (t) => {
   rimraf.sync(db2Path)
   t.pass('delete db2 folder to start clean')
 
-  const keys = ssbKeys.loadOrCreateSync(path.join(dir, 'secret'))
   const sbot = SecretStack({ appKey: caps.shs })
     .use(require('ssb-db'))
     .use(require('../migrate'))
@@ -126,7 +131,6 @@ test('migrate (alone)', async (t) => {
   rimraf.sync(db2Path)
   t.pass('delete db2 folder to start clean')
 
-  const keys = ssbKeys.loadOrCreateSync(path.join(dir, 'secret'))
   const sbot = SecretStack({ appKey: caps.shs })
     .use(require('../migrate'))
     .call(null, { keys, path: dir })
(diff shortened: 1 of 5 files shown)


staltz:
@arj03 I pushed the implementation I have so far, and then I'm stopping my work day today. The tests are going to fail (just one), but **if** the benchmarks ran, you'd see a much smaller number for `reboot and key one again`. Maybe you can run it locally if you're curious.

arj03:
Cool thanks. I'm going home now. Will sneak a peak at this tonight though :-)

staltz:
@arj03 This test fails, isn't it supposed to be `t.equal(msg.value.previous, null)`? Because in the previous test, `delete all`, the state of that feed should now be removed from the base index

https://github.com/ssb-ngi-pointer/ssb-db2/blob/5b538159676997de74bc458c655ad145e1310b81/test/basic.js#L188

arj03:
Some numbers from my full db (1.5 mill messages) machine: 6 seconds index create, 92M size.

staltz:
Ouch, 92MB is a lot. I forgot to check that up. Hmm, I wonder if I could do a prefix technique...

arj03:
The keys map prefix index is 34mb in comparison. I can work on updating ssb-friends to the new api, then I can do a better test of performance.

arj03:
This PR decreases the memory usage by 75mb roughly.

staltz:
:arrow_up: Check the benchmark `reboot and key one again`

arj03:
Yeah saw that one locally yesterday as well. It is quite an heavy index to load.

staltz:
Pushed a commit to cut the prefix and suffix in a key. Let's see the benchmarks

staltz:
Lol, basically no change

staltz:
@arj03 We could merge this (for experimentation) or I could install ssb-db2 in Manyverse referencing the `keys-stuff` branch directly. What do you think?

arj03:
Yeah and actually locally here rebuilding the database went from 6s to almost 9. I guess the overhead of creating an extra string is significant.

staltz:
>  went from 6s to almost 9

Wow. Let's see if the CI benchmarks say a similar story too.

staltz:
I'll do this: revert back to full key string, *not merge* this PR (yet), test how it goes in manyverse, and then we can talk about it again.

staltz:
Yeah, tested on mobile, and startup is 3 sec faster (down from 8s to 5s).

Indexes folder looks like this:

```
$ du -h

2.4M	./contacts
1.4M	./aboutSelf
2.6M	./base
4.6M	./fullMentions
17M	./ebt
57M	./keys
```

The original `key__map.32prefixmap` was `21M`.

arj03:
That is good news!

staltz:
Yeah, I checked the old folders on mobile, backlinks takes 262MB, `links` takes 330MB, ssb-threads took 129MB.

The new keys index is not the tiniest, but it's certainly an improvement over db1. Later if we find a smarter way, we can bump the version number for this index. I think this PR would be nice to merge for manyverse, but what's your thoughts on the browser use case?

arj03:
I'll test it later this afternoon and let you know

arj03:
Works great in browser as well. Lets get this merged :-)

arj03 (approved):

PR #180, opened 2021-02-15
add test for base status always in getStatus()

staltz:
1st test, 2nd fix


staltz:
Context: I need this kind of information to more accurately calculate progress percentages

Problem: base index is a special one, it doesn't go through `registerIndex`

Solution: call this ad-hoc method

arj03 (approved):

PR #181, opened 2021-02-15
before deleting index item, must flush batch

staltz:
The tests are wrong in one spot, see this comment: https://github.com/ssb-ngi-pointer/ssb-db2/pull/179#issuecomment-779356642

Also, besides the tests being wrong, also so was the implementation of delete. I don't have proof in the tests, but you can easily confirm this by yourself by putting some console.logs here and there: the `level.del` happens **1st** and **then** our batch system kicks in, **re-adding** the item that should be deleted.

This PR adds a commit that fixes the delete, to flush first. This will cause tests to fail. Then, I'll fix the test.

### DIFF
--- a/indexes/base.js
+++ b/indexes/base.js
@@ -73,8 +73,11 @@ module.exports = class BaseIndex extends Plugin {
   }
 
   removeFeedFromLatest(feedId) {
-    this.level.del(feedId, (err) => {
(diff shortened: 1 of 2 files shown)


arj03 (approved):

arj03:
nice find, must have been a mind fuck with the test :(

PR #184, opened 2021-02-16
refactor base index to use registerIndex()

staltz:
So that it uses `registerIndex()`, and then we can be more confident about putting more logic in `registerIndex()` and knowing it will apply to base too.


arj03 (approved):
All your base are belong to us!

PR #186, opened 2021-02-16
add sbot.keys in compat

staltz:
This was missing, it's needed in https://github.com/staltz/ssb-lan/blob/230daca3df3d00742d31938b398dcb9a212da031/src/index.ts#L95 and in db1 it's defined here https://github.com/ssbc/ssb-db/blob/8312ebe7df1b2b4e3282f75c73c753d8bc425dae/index.js#L122

### DIFF
--- a/compat/db.js
+++ b/compat/db.js
@@ -12,6 +12,7 @@ exports.init = function (sbot, config) {
   sbot.publish = sbot.db.publish
   sbot.whoami = () => ({ id: sbot.id })
   sbot.ready = () => true
+  sbot.keys = config.keys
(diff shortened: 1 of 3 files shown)


arj03 (approved):

arj03:
The keys is actually generally pretty useful by itself. Might want to put it into db directly. But lets just merge this.

staltz:
Yeah it's useful, but it's always there in `config.keys` for any secret-stack plugin that really needs it. I would think that the database module `ssb.db` is not strictly responsible for holding cryptographic keys, so I'm happy if it would be in `config.keys` for people not installing `compat`

PR #191, opened 2021-02-17
adjust maxCpu maxPause=180ms

staltz:
I ran some experiments, and found this sweet spot:

| initial sync duration | maxCpu    | UI responsiveness |
|------|-----------|-------------------|
| 305s | 86 (maxPause=300) | good              |
| 299s | 90 (maxPause=300) | good              |
| 288s | 92 (maxPause=300) | good              |
| **222s** | **92 (maxPause=180)** | **good**              |
| 158s | 92 (maxPause=90)  | a bit bad         |
| 281s | 95 (maxPause=300) | a bit bad         |
| 280s | 99 (maxPause=300) | bad               |
| 126s | Infinity          | bad   |

### DIFF
--- a/log.js
+++ b/log.js
@@ -51,7 +51,7 @@ module.exports = function (dir, config, privateIndex) {
   log.stream = function (opts) {
     const shouldDecrypt = opts.decrypt === false ? false : true
     const tooHot = config.db2.maxCpu
-      ? TooHot({ ceiling: config.db2.maxCpu, wait: 90, maxPause: 300 })
+      ? TooHot({ ceiling: config.db2.maxCpu, wait: 90, maxPause: 180 })
       : () => false
     const s = originalStream(opts)
     const originalPipe = s.pipe.bind(s)


staltz:
Actually I'm thinking that I should make this configurable instead of hard-coded.

arj03:
Yeah, would be good if it was configurable. How is indexing after the async-log fix and with this? Is it faster or about the same?

staltz:
I think it's the roughly same, but I'm reindexing my production mobile account to make an accurate measurement.

staltz:
Initial indexing took 20min this time, down from 25min before. But it's not showing votes at all :(

staltz:
> But it's not showing votes at all :(

Oh, now it's showing. It's funny, like there was some hiccup in the JIT detection. After initial indexing, it didn't show votes. Then I opened the app again, didn't show votes again. Third time I opened, then it showed the votes.

Maybe what happened was that everything else indexed just fine, enough to display content. But the threads were shown while votes was still indexing. And because I killed the app, the indexing of votes restarted from zero (we don't yet have saving of JIT indexes every X seconds, I should do that PR before putting db2 manyverse in production). @arj03

PR #192, opened 2021-02-17
allow all too-hot params to be configured

staltz:
This PR is better than #191

### DIFF
--- a/README.md
+++ b/README.md
@@ -370,6 +370,18 @@ const config = {
      * Default: Infinity
      */
     maxCpu: 85,
+
+    /** This applies only if `maxCpu` is defined.
+     * See `maxPause` in the module `too-hot`, for its definition.
+     * Default: 300
+     */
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -370,6 +370,18 @@ const config = {
      * Default: Infinity
      */
     maxCpu: 85,
+
+    /** This applies only if `maxCpu` is defined.
+     * See `maxPause` in the module `too-hot`, for its definition.
+     * Default: 300
```
arj03, line 376: 180

arj03: or is it the default of too-hot?

staltz: Check line 12 in log.js, I kept it at 300. That's because 300 is good in CI. 180 is good for mobile.

arj03: Ah right, this is from the README :)

arj03 (approved):

PR #203, opened 2021-02-23
implement config dangerouslyKillFlumeWhenMigrated

staltz:
@arj03 This one should be reviewed with triple caution. The long name was chosen like that to scare developers even further.


staltz:
Clearly something wrong with the tests too.

staltz:
Tests were just flaky. I'll make another PR to reduce flakiness.

staltz:
Improved the dangerouslyKillFlumeWhenMigrated test a bit.

arj03:
I'll do a proper review of this tomorrow with a clear head :)

arj03:
I need to understand exactly how you are migrating things. Do you have db1 still enabled? Or have you just enabled the compat plugins and then you let it migrate? Meaning things like EBT will write messages into these queues?

migrate.js
```diff
@@ -259,32 +289,23 @@ exports.init = function init(sbot, config) {
...
 
-        // Setup migration of live new msgs identified on the old log
-        oldLog.newMsgObv((msg) => {
-          const guard = guardAgainstDecryptedMsg(msg)
-          if (guard) throw guard
-
-          writeToNewLog(toBIPF(msg), () => {
-            liveMsgCount++
-            if (liveMsgCount % 100 === 0) emitProgressEvent()
+        if (config.db2.dangerouslyKillFlumeWhenMigrated) {
```
staltz, line 300: One thing that I wanted to do here was `oldLog.close` but I noticed such method does not exist. :eyes: I think we're safe enough from the possibility of FlumelogOffset instance being used, but ideally it would be good to nuke that possibility.

staltz:
> I need to understand exactly how you are migrating things. Do you have db1 still enabled? Or have you just enabled the compat plugins and then you let it migrate? Meaning things like EBT will write messages into these queues?

db1 is not installed at all. And yes I have the compat plugins. It's a good question, though, what happens if it is installed.

EBT will write messages into the queues. I'm doing a test right now, by the way:

- Flume exists, no DB2 folder yet
- Phone connected to the internet (thus *can* connect to other peers)
- Migrate begins, will dangerouslyKillFlumeWhenMigrated
- I'm expecting that EBT writes will get queued and applied only after migration is done

staltz:
> I'm doing a test right now, by the way:

Well unfortunately my test doesn't trigger the EBT writes at that moment because in Manyverse it only starts ssb-conn once the first thread has been shown on the screen, which is after migrate (and all db2 indexes) are done.

arj03:
Hmm ok. I would have recommended to do that anyway because I wouldn't be 100% sure that the EBT works. And what happens if you close the migration in the middle. All the queue stuff is only in memory, so you might end up in a wierd state where the other end thinks you have received something you might not. Just being super paranoid here :)

staltz:
@arj03 New commit pushed

db.js
```diff
@@ -143,7 +158,9 @@ exports.init = function (sbot, config) {
   }
 
   function add(msg, cb) {
-    const guard = guardAgainstDuplicateLogs('add()')
+    var guard = guardAgainstMigrationDangers('add()')
+    if (guard) return cb(guard)
+    var guard = guardAgainstDuplicateLogs('add()')
```
staltz, line 163: @arj03 Oh wait, I'll merge these two checks into one function :)

db.js
```diff
@@ -270,8 +252,9 @@ exports.init = function (sbot, config) {
   }
 
   function deleteFeed(feedId, cb) {
-    if (waitingForMigrated()) return queues.deleteFeed.push([feedId, cb])
-    const guard = guardAgainstDuplicateLogs('deleteFeed()')
+    var guard = guardAgainstMigrationDangers('deleteFeed()')
+    if (guard) return cb(guard)
+    var guard = guardAgainstDuplicateLogs('deleteFeed()')
     if (guard) return cb(guard)
```
arj03, line 258: looks like this can be refactored :-)

staltz:
Ok here we go.

staltz:
Nope, gotta fix the test assertion.

staltz:
Now should be good.

arj03:
I think we are good now. I was staring at the drainAborter = null line, but can see if was just moved from old code

arj03 (approved):

staltz:
Thanks for reviewing! I have another simple PR coming up, about optimizing emitProgressEvent

PR #206, opened 2021-02-25
optimize how migrate progress events are done

staltz:
Two ideas:

- In non-live mode, when the setInterval triggers, only call the `sbot.emit` if the number is different (I detected in production that sometimes the number is the same, I guess it was busy with indexing or something)
- In live mode, don't emit every 100th time, instead, use an interval timer too


arj03 (approved):

staltz:
Thanks

Issue #207, opened 2021-03-02
Some issues with aboutSelf


staltz:
I'm pretty sure (and @powersource also found some cases) where the user's about is a very old one, definitely not the latest self-assigned, e.g. `andrestaltz` (desktop) below, the profile pic is super old:

![photo_2021-03-02_16-06-22](https://user-images.githubusercontent.com/90512/109660210-425f6100-7b71-11eb-8d50-57de3f298bec.jpg)

arj03:
I can have a go at debugging that if you want?

staltz:
Oh, good! Thanks

arj03:
Just tested on my main log and I get the correct picture. It finished indexing right?

Powersource:
Yep issue persists after several days of using the app

staltz:
> Just tested on my main log and I get the correct picture. It finished indexing right?

I think it finished indexing because I see from logs that db2 loaded such and such plugins, and all of them including aboutSelf have the same offset and processed number.

staltz:
I'll try to reproduce on my desktop.

staltz:
Can't reproduce on desktop, I'm now trying on mobile to turn off the "cachedAboutSelf" plugin, let's see what that does.

Nope, didn't make a difference.

Let's see if deleting `aboutSelf` and letting it rebuild helps.

Nope, still giving old abouts. This is mysterious...

arj03:
At least it is reproducible on mobile ;-)

arj03:
I ran this to make sure that I had all the about messages, maybe that tells you something:

```
sbot.db.query(
  and(
   type('about'),
   author('@QlCTpvY7p9ty2yOFrv1WU1AE88aoQc4Y7wYal7PFc+w=.ed25519')
  ),
  descending(),
  toCallback((err, msgs) => {
    console.log(msgs.filter(m => m.value.content.about == m.value.author).map(m => m.value.content))
    sbot.close()
  })
)
```

staltz:
I put custom logs in and it seems that the abouts are all correctly there, they just come in the wrong order which means that one of the older ones is the last one to call `updateProfileData`, and that's the last thing that remains there in leveldb.

Now debugging why that happens...

staltz:
Nothing seems wrong with the order of offset and seq of self-assigned abouts of `@QlCT...` on my phone, see:

```
88401 172
88818 173
101175 198
1541622 2666
123419498 170755
166547364 224141
[...]

PR #208, opened 2021-03-03
recalculate ssb-validate state when migrate ends

staltz:
Tested on my manyverse and didn't get #207 symptoms anymore. 

Ideally, it would be good to have tests for this, but it's quite a lot of work to setup everything to reproduce it (would require migrating a fixture, then replicating with some peers which have more data on that fixture). Doable, but I'm on a deadline.

Some more comments inline

### DIFF
--- a/compat/ebt.js
+++ b/compat/ebt.js
@@ -1,4 +1,5 @@
 const EBTIndex = require('../indexes/ebt')
+const { onceWhen } = require('../utils')
 
 exports.init = function (sbot, config) {
   sbot.db.registerIndex(EBTIndex)
@@ -7,15 +8,23 @@ exports.init = function (sbot, config) {
   sbot.getAtSequence = ebtIndex.getMessageFromAuthorSequence.bind(ebtIndex)
   sbot.add = sbot.db.add
   sbot.getVectorClock = function (cb) {
-    sbot.db.getAllLatest((err, last) => {
-      if (err) return cb(err)
+    onceWhen(
+      sbot.db2migrate && sbot.db2migrate.synchronized,
+      (isSynced) => isSynced,
+      () => {
+        sbot.db.onDrain('base', () => {
+          sbot.db.getAllLatest((err, last) => {
+            if (err) return cb(err)
 
-      const clock = {}
-      for (const k in last) {
-        clock[k] = last[k].sequence
+            const clock = {}
+            for (const k in last) {
+              clock[k] = last[k].sequence
+            }
+
+            cb(null, clock)
+          })
+        })
       }
-
-      cb(null, clock)
-    })
+    )
   }
 }
--- a/db.js
+++ b/db.js
@@ -7,9 +7,9 @@ const jitdbOperators = require('jitdb/operators')
 const operators = require('./operators')
 const JITDb = require('jitdb')
 const Debug = require('debug')
-const DeferredPromise = require('p-defer')
 
 const { indexesPath } = require('./defaults')
+const { onceWhen } = require('./utils')
 const Log = require('./log')
 const Status = require('./status')
 const makeBaseIndex = require('./indexes/base')
@@ -57,7 +57,7 @@ exports.init = function (sbot, config) {
   const debug = Debug('ssb:db2')
   const post = Obv()
   const hmac_key = null
-  const stateFeedsReady = DeferredPromise()
+  const stateFeedsReady = Obv().set(false)
   let state = validate.initial()
 
   sbot.close.hook(function (fn, args) {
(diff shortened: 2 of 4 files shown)


db.js
```diff
@@ -301,25 +320,28 @@ exports.init = function (sbot, config) {
...
-          const remove = index.offset(() => {
-            if (index.offset.value === log.since.value) {
-              remove()
-              status.updateIndex(indexName, index.offset.value)
-              cb()
-            }
-          })
-        }
+    // setTimeout to make sure extra indexes from secret-stack are also included
+    setTimeout(() => {
```
staltz, line 324: This new setTimeout was important to fix the NaN offset issue. Some onDrain's out there, such as the newly added one in EBT compat, would cause the trigger of updateIndexes() *before* all the extra level plugins had the chance to get registered, which means that the `Math.min` would operate on `undefined` values, and that would result in `NaN` and then it would be a log.stream `gt: NaN`. So we need to delay this just a bit.

migrate.js
```diff
@@ -221,6 +221,7 @@ exports.init = function init(sbot, config) {
     started = true
     debug('started')
 
+    if (sbot.db) sbot.db.setStateFeedsReady('migrating')
```
staltz, line 224: This `'migrating'` is a trick to not let `loadStateFeeds` flag "true" for stateFeedsReady. Basically db.js might have run `loadStateFeeds` for other reasons than migration, but we want to prevent db.js from declaring "true" on stateFeedsReady, because migration will do that (see line 320 in migrate.js).

utils.js
```diff
@@ -0,0 +1,21 @@
+/**
+ * Obv utility to run the `cb` once, as soon as the condition given by
+ * `filter` is true.
+ */
+function onceWhen(obv, filter, cb) {
```
staltz, line 5: There's another discussion to be had one day, but it's not important for this PR, which is: should we keep using Obv? I find it rather weak in API compared to other observable libraries out there, and there are certainly good ones that have decent API, are very minimal in bundle size, and performant. But for now, let's just keep using Obv.

compat/ebt.js
```diff
@@ -7,15 +8,23 @@ exports.init = function (sbot, config) {
   sbot.getAtSequence = ebtIndex.getMessageFromAuthorSequence.bind(ebtIndex)
   sbot.add = sbot.db.add
   sbot.getVectorClock = function (cb) {
-    sbot.db.getAllLatest((err, last) => {
-      if (err) return cb(err)
+    onceWhen(
+      sbot.db2migrate && sbot.db2migrate.synchronized,
```
staltz, line 12: If this first argument in `onceWhen` is falsy, then `onceWhen` will just call the callback. So no worry if you're not using sbot.db2migrate

db.js
staltz: Nothing new in the body, just a setTimeout and a re-indent. Makes the diff look scary

utils.js
```diff
@@ -0,0 +1,21 @@
...
+  if (!obv) return cb()
+  let answered = false
+  let remove
+  remove = obv((x) => {
+    if (!filter(x)) return
+    if (answered) {
+      if (remove) remove(), (remove = null)
```
arj03, line 12: I'm not used to this. Can we do a { and 2 lines instead?

utils.js
arj03: I wouldn't mind changing to something else than Obz. Fixed a pretty bad bug in it in the beginning of db2 that has been there for ages. I think it's one of those, build by dominic and thus used a lot in ssbc, but not much elsewhere.

migrate.js
arj03: I'm wondering if there is no concurrency issues here where migrate sets this to migrating first, and then db.js sets it to true afterwards?

migrate.js
```diff
@@ -313,10 +314,20 @@ exports.init = function init(sbot, config) {
...
+                oldLogExists.set(false)
+                doneMigrating()
+              })
+            } else {
+              oldLogExists.set(false)
+              doneMigrating()
+            }
           })
         } else {
+          if (sbot.db) sbot.db.loadStateFeeds()
```
arj03, line 330: Shouldn't this line also do a `sbot.db.setStateFeedsReady(true)` like above. For me this looks like the case where you are migrating but don't have kill log enabled?

staltz: Oops, that's correct, will fix.

arj03:
Argh github is buggy, doubled posted my comments but then when I deleted double deleted :(

utils.js
arj03: Since my other response was eaten by /dev/null. Yes

db.js
```diff
@@ -148,15 +159,19 @@ exports.init = function (sbot, config) {
...
     if (guard) return cb(guard)
 
-    stateFeedsReady.promise.then(() => {
-      try {
-        state = validate.append(state, hmac_key, msg)
-        if (state.error) return cb(state.error)
-        rawAdd(msg, true, cb)
-      } catch (ex) {
-        return cb(ex)
+    onceWhen(
```
arj03, line 162: I have a little trouble wrapping my head around how this functions when migrate is not enabled? Can you walk me through how it works in onceWhen? Like is there a fast path it will take then?

arj03:
I hope you got my replies in email at least ;-) Very happy it is running well in real world testing. Overall I like this change. It looks solid.

Two things left otherwise this looks good:

 - sbot.db.setStateFeedsReady('migrating'). This one about concurrency
 - and the onceWhen, just need to understand exactly how that works

db.js
staltz: I think this code is not related to migrate in any way, when migrate is turned off, this `onceWhen` will just listen to the `stateFeedsReady` Obv (it used to be a promise, but we need to bring it back to `false` (or "migrating") sometimes, so I made it an Obv), and when that Obv finally satisfies the second argument `ready => ready === true`, then the cb is called and it will never be called again (because it's "once").

staltz:
> I hope you got my replies in email at least

Sorry, I don't have github emails enabled :(



>     * sbot.db.setStateFeedsReady('migrating'). This one about concurrency

I didn't see any comment about this one. Was it deleted by GH?

arj03:
> > ```
> > * sbot.db.setStateFeedsReady('migrating'). This one about concurrency
> > ```

My question was, are we sure there is not a concurrency problem where first the migrate set this to 'migrating' and then db set it to true after loading lets say half of the migrated data.

staltz:
Right, so this is handled in db.js line 93, db.js will only set ready to `true` if it was false before. `"migrating"` is not falsy, so it doesn't touch it.

arj03:
Thanks makes sense

db.js
arj03: I ran a test locally, mostly just so I understand how this works so mostly the calls just runs through the Obz. Then overhead doesn't seem quite minimal. I noticed there was no test of the performance of adding a bunch of messages so I tried that locally. Will do a separate PR for that. The result was the same though for this and master.

arj03 (approved):
Very happy with this. Thanks for slaying the hopefully last :dragon_face:

staltz:
The Last Dragon, could be the name of a movie

PR #209, opened 2021-03-03
Add test for adding a bunch of messages

arj03:
I noticed we didn't have a benchmark of simply adding a bunch of messages. This is useful for testing stuff like how fast can we add 1000 messages coming from EBT.

### DIFF
--- a/benchmark/index.js
+++ b/benchmark/index.js
@@ -9,6 +9,8 @@ const caps = require('ssb-caps')
 const ssbKeys = require('ssb-keys')
 const multicb = require('multicb')
 const pull = require('pull-stream')
+const asyncFilter = require('pull-async-filter')
+const validate = require('ssb-validate')
 const fromEvent = require('pull-stream-util/from-event')
 const DeferredPromise = require('p-defer')
 const trammel = require('trammel')
@@ -29,10 +31,13 @@ const {
 } = require('../operators')
 
 const dir = '/tmp/ssb-db2-benchmark'
+const dirAdd = '/tmp/ssb-db2-benchmark-add'
 const oldLogPath = path.join(dir, 'flume', 'log.offset')
 const db2Path = path.join(dir, 'db2')
 const indexesPath = path.join(dir, 'db2', 'indexes')
 const reportPath = path.join(dir, 'benchmark.md')
+
+rimraf.sync(dirAdd)
 
 const skipCreate = process.argv[2] === 'noCreate'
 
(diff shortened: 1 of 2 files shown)


benchmark/index.js
```diff
@@ -87,6 +92,42 @@ test('setup', (t) => {
   t.end()
 })
 
+test('add a bunch of messages', async (t) => {
```
arj03, line 95: try changing this test to 'add 1000 messages' and see what happens :octopus:

benchmark/index.js
```diff
@@ -119,6 +160,7 @@ test('migrate (+db1)', async (t) => {
       fs.appendFileSync(reportPath, `| Migrate (+db1) | ${duration}ms |\n`)
       await sleep(2000) // wait for new log FS writes to finalize
       sbot.close(() => {
+        t.end()
```
arj03, line 163: I noticed quite a few of these had this missing. Seemed like a good idea to have, or is there an implicit t.end?

arj03:
For reference locally add 1000 elements is 300ms.

benchmark/index.js
staltz: Oh, sorry that this is a bit mysterious, but we're using promises in these `async-await` tests, which means we don't need to do a t.end(), see https://github.com/substack/tape#testname-opts-cb (search "promise")

arj03: Ok, will remove those then

benchmark/index.js
```diff
@@ -87,6 +92,42 @@ test('setup', (t) => {
...
+  let state = validate.initial()
+  for (var i = 0; i < 1000; ++i) {
+    state = validate.appendNew(state, null, keys, { type: 'tick', count: i }, Date.now())
+  }
+
+  const ended = DeferredPromise()
+  const start = Date.now()
+
+  pull(
+    pull.values(state.queue.map(x => x.value)),
```
staltz, line 109: To measure only sbot.db.add, could we move this array map to up there before the `start`? Like `const vals = state.queue.map(x => x.value)`. It's probably not too many milliseconds, but better to be sure

benchmark/index.js
```diff
@@ -87,6 +92,42 @@ test('setup', (t) => {
...
+    asyncFilter(sbot.db.add),
+    pull.collect((err) => {
+      if (err) t.fail(err)
+
+      const duration = Date.now() - start
```
staltz, line 114: The other benchmarks have the `duration` declaration before the `if (err) t.fail(err)`, to avoid measuring the `if`. I know, the `if` lasts 1 nanosecond maximum, but just for consistency with the other code, and since you're changing code here, might as well just move it

staltz:
Really good idea to measure this!

staltz:
What's your interpretation, is 1000 msgs written in 300ms fast or slow? I don't have an intuition for this one

staltz:
There's one thing I'm really puzzled about:

| Part | Duration |
|---|---|
| add 1000 elements | 2059ms |
| Migrate (alone) | 3364ms |

Migrate is writing 100k msgs, so it should last 205900ms if we take the `add 1000 elements` as a reference. What's going on here? :|

staltz:
Oh, validate.

staltz:
> Migrate is writing 100k msgs, so it should last 205900ms if we take the `add 1000 elements` as a reference. What's going on here? :|

Just for my own reference for the future, this means validate makes `add` 61x slower.

arj03:
Besides the reason mentioned in the PR message, I wanted to add this benchmark because if we come up with a better solution than the Obv in add, then we can measure the performance as well, I'm pretty sure it's validate that is slow though.

staltz (approved):

PR #212, opened 2021-03-11
Auto encrypted publish when recps is set

### LINKED ISSUE #211: Publish doesn't encrypt messages by default
Memo raised the issue here (%u0MgEX7mwPmJHcikBBOchPHIy0lkyfPS/7C/cM6YxwU=.sha256) that the function publish in db2 works differently than db1 for private messages in that it expects the message to already be [encrypted](https://github.com/ssb-ngi-pointer/ssb-db2/blob/c0d78cfe910d520f685a7eceb602bbf566a59220/test/base-index.js#L98).

Some possible remedies:
 - Auto encrypt messages if recps is set
 - Document that publish expects a private message to already be encrypted and fail if it contains recps

arj03:
To fix #211. Current behaviour can be a bit of a gotcha when coming from db1. First commit is just a failing test.

### DIFF
--- a/db.js
+++ b/db.js
@@ -219,10 +219,7 @@ exports.init = function (sbot, config) {
       (ready) => ready === true,
       () => {
         if (msg.recps) {
-          msg = ssbKeys.box(
-            msg,
-            msg.recps.map((x) => x.substr(1))
-          )
+          msg = ssbKeys.box(msg, msg.recps)
         }
 
         state.queue = []
--- a/db.js
+++ b/db.js
@@ -1,4 +1,5 @@
 const push = require('push-stream')
+const ssbKeys = require('ssb-keys')
 const hash = require('ssb-keys/util').hash
 const validate = require('ssb-validate')
 const Obv = require('obz')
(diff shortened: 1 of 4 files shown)


arj03:
Tests are being annoying, seems like the compile of sodium takes too long time.

db.js
```diff
@@ -217,8 +218,16 @@ exports.init = function (sbot, config) {
       stateFeedsReady,
       (ready) => ready === true,
       () => {
+        if (msg.recps) {
+          msg = ssbKeys.box(
+            msg,
+            msg.recps.map((x) => x.substr(1))
```
mixmix, line 224: This only supports classic encryption. A good start but we need an interface like add Boxer and addUnboxer right?

arj03: Yes, but I would rather do that in https://github.com/ssb-ngi-pointer/ssb-db2/pull/90. Would like https://github.com/ssb-ngi-pointer/ssb-fixtures/pull/2 to be in a bit better shape before really looking at that again though.

staltz: Hmm, I was looking at that `x.substr(1)` and got confused. Do we need to remove the sigil `@`?

According to ssb-keys implementation, yes: 

https://github.com/ssb-js/ssb-keys/blob/98569aa430e344b3ed144fcebf5d2755f7355bf6/index.js#L139

But according to how Patchwork does it right now, it checks if it's a `Ref.isFeed()`:

https://github.com/ssbc/patchwork/blob/ae29bd7f91e16deac4bd88b099175fa85a419815/lib/depject/sbot.js#L172-L174

How does Patchwork even work?

arj03: @staltz https://github.com/ssb-js/ssb-keys/pull/92 ;-)

arj03: Given that it also works that way I'm open to maybe doing this without the substr. What do you think?

staltz: Yeah this substr here seems not necessary if ssb-keys does it in utils.toBuffer

staltz (approved):
:rocket:

PR #215, opened 2021-03-15
clearInterval from migrate plugin

staltz:


staltz:
This fixed the big memory numbers. Turned out it was a benchmark artifact. All those `setIntervals` stayed up, when they should have died with each sbot.close.

Although this is great to merge, it doesn't explain the memory leaks in manyverse in production because this "migrate live" code doesn't run at all in manyverse (because we use dangerously kill flume).

arj03 (approved):

PR #217, opened 2021-03-18
Fix error callback issue

Barbarrosa:
I ran ESLint on this repo and spotted a small bug. Here's the fix.


arj03:
Fix looks good. Thanks :) Will just rerun tests as it seems it had a a minor issue in benchmark that should be unrelated to this PR.

arj03:
Now I see. Thanks again.

PR #218, opened 2021-03-29
Private 3

arj03:
Continuing https://github.com/ssb-ngi-pointer/ssb-db2/pull/90. We might need quite a large part of this if we do feedless identities so I wanted to get that branch updated to latest master. Turns there there was a bunch of more things we need to fix:

 - For encrypted we don't store seq, only offset and that doesn't work well with full-mentions and keys index
 - Tribes 2 has a wierd edge case where it doesn't close properly

### DIFF
--- a/db.js
+++ b/db.js
@@ -51,7 +51,7 @@ exports.init = function (sbot, config) {
   config.db2 = config.db2 || {}
   const indexes = {}
   const dir = config.path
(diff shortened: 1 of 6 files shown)


mixmix:
Arg, I tried to put good close hooks in there.

staltz:
Can we delete this branch?

PR #219, opened 2021-04-01
Bump bipf

arj03:
Bipf was locked at 1.4. I have been doing some work on jitdb and noticed that db2 will not use 1.5 even if jitdb will because of the ~. Test parses and they also parse in jitdb.


staltz (approved):

staltz:
This PR could have been named "Bumpf"

PR #220, opened 2021-04-03
use jitdb@3 and where() operator

staltz:
Follow up to https://github.com/ssb-ngi-pointer/jitdb/pull/141. Also a **breaking change**.

### DIFF
--- a/db.js
+++ b/db.js
@@ -17,7 +17,15 @@ const makeBaseIndex = require('./indexes/base')
 const KeysIndex = require('./indexes/keys')
 const PrivateIndex = require('./indexes/private')
 
-const { and, fromDB, key, author, deferred, toCallback, asOffsets } = operators
+const {
+  where,
+  fromDB,
+  key,
+  author,
+  deferred,
+  toCallback,
+  asOffsets,
+} = operators
 
 function getId(msg) {
   return '%' + hash(JSON.stringify(msg, null, 2))
@@ -126,7 +134,7 @@ exports.init = function (sbot, config) {
 
   function getHelper(id, onlyValue, cb) {
     self.query(
-      and(key(id)),
+      where(key(id)),
       toCallback((err, results) => {
         if (err) return cb(err)
         else if (results.length)
@@ -236,7 +244,7 @@ exports.init = function (sbot, config) {
     if (guard) return cb(guard)
 
     self.query(
-      and(key(msgId)),
+      where(key(msgId)),
       asOffsets(),
       toCallback((err, results) => {
         if (err) return cb(err)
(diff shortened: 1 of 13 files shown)


staltz:
No idea why tests keep failing, they are passing locally.

arj03:
I also can't reproduce the error locally. Seems to be a sorting problem in jitdb. I'm tempted to rerun the tests here, but I'm not sure if that is going to change anything because both node 10 and 12 failed.

db.js
```diff
@@ -392,18 +400,12 @@ exports.init = function (sbot, config) {
...
-    if (!first.meta) {
+    if (first.meta) {
+      return jitdbOperators.query(first, where(waitUntilReady), ...rest)
+    } else {
       const ops = fromDB(jitdb)
       ops.meta.db = this
-      return jitdbOperators.query(ops, and(waitUntilReady, first), ...rest)
-    } else {
-      return jitdbOperators.query(first, and(waitUntilReady), ...rest)
+      return jitdbOperators.query(ops, where(waitUntilReady), first, ...rest)
```
arj03, line 408: At first reading this, I thought this needed a `where(and(waitUntilReady, first)` but now I see that doesn't work.

indexes/full-mentions.js
```diff
@@ -64,7 +64,7 @@ module.exports = class FullMentions extends Plugin {
             pl.read(this.level, Object.assign({}, opts, { live, old: false })),
             pull.map(parseInt10)
           )
-          cb(null, or(seqs(seqArr.map(parseInt10)), liveSeqs(ps))())
```
arj03, line 67: this is a nice improvement!

test/operators.js
```diff
@@ -199,36 +199,7 @@ test('execute and(hasRoot(msgkey))', (t) => {
         t.error(err, 'no err')
 
         db.query(
```
arj03, line 201: finally found something to nitpick, the test needs to be called `execute hasRoot(msgkey)` ;-)

arj03: Pushed up a change for this to try and run the tests again :)

arj03:
Well at least the failure is reproducible.

db.js
staltz: Yeah, `first` could be `live` and `live` doesn't work inside `and` anymore (it used to).

arj03:
This is really strange. It seems like there actually also was the same bug in the commit that bumped the version to 1.18.7. I just don't see why anything in bipf 1.5 should have triggered this problem. I'll try a debug branch.

arj03 (approved):

arj03:
The error is unrelated to this PR. The error is randomly triggered because we basically add messages from the future and jitdb uses [min of arrival, declared](https://github.com/ssb-ngi-pointer/jitdb/blob/2ba4c90b558f5034a9e82a01e8f941a89d253bcc/index.js#L253). I'll fix the tests.

staltz:
Great! Let's merge

PR #222, opened 2021-04-05
experiment fast-json-stringify

staltz:
I don't think this PR gives any benefit, but it was an old branch lying around locally and I thought of sharing it here in case it gives anyone some insight into something new.


staltz:
Bye bye old PR

PR #225, opened 2021-04-20
Faster validate on add

arj03:
Was testing rust validation and discovered that we using the ssb-validate in a suboptimal way. Basically recalculating the hash of the messages.

Before:

```
 add a bunch of messages

    ✔ duration: 361ms
```

After:

```
 add a bunch of messages

    ✔ duration: 292ms
```

bunch=1000.


arj03:
@mycognosist

staltz (approved):
Dude, the git diff is net negative lines-of-code, all tests pass, and it's faster. Instamerge.

PR #228, opened 2021-05-11
Use ssb-validate2-rsjs

arj03:
WIP using rust validation instead of ssb-validate.

### DIFF
--- a/db.js
+++ b/db.js
@@ -157,6 +157,28 @@ exports.init = function (sbot, config) {
     state[kvt.value.author] = kvt
   }
 
+  function addOOOBatch(msgs, cb) {
+    const guard = guardAgainstDuplicateLogs('addOOOBatch()')
+    if (guard) return cb(guard)
+
+    onceWhen(
+      stateFeedsReady,
+      (ready) => ready === true,
+      () => {
+        const kvts = msgs.map((m) => toKeyValueTimestamp(m))
+        const err = validate.validateOOOBatch(kvts)
+        if (err) return cb(err)
+
+        const done = multicb({ pluck: 1 })
+        kvts.forEach((kvt) => {
+          log.add(kvt.key, kvt.value, done())
+        })
+
+        done(cb)
+      }
+    )
+  }
+
   function addBatch(msgs, cb) {
     const guard = guardAgainstDuplicateLogs('addBatch()')
     if (guard) return cb(guard)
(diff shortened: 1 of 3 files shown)


db.js
```diff
@@ -85,17 +89,20 @@ exports.init = function (sbot, config) {
...
-            queue: [],
-          }
-        }
-        debug('getAllLatest is done setting up initial validate state')
-        if (!stateFeedsReady.value) stateFeedsReady.set(true)
-        if (cb) cb()
+
+        const done = multicb({ pluck: 1 })
+
+        for (const k in last) getMsg(last[k].id, done())
```
staltz, line 95: It's a bit unfortunate we have to do this lookup, since the base index is supposed to support ssb-validate, what if we redesign base index and ebt index so that they support loading up the feed states more directly?

As a reminder, currently:

- Base index: `authorId => latestMsg { key, sequence, timestamp }`
- EBT index: `[authorId, sequence] => offset`

What if the base index is just `authorId => latestMsg { offset }`? Then we can just do one async-aol fetch of that offset, and then we get the full msg which has the `sequence` and `timestamp` and `key` (obviously, it has everything).

arj03: We can change base index to that. Bendy butt validation would also like to have access to the full previous message. First I'd like to know what the difference is on startup, I'll test that. This shouldn't have an impact on EBT as that is a author, seq per message, so I don't think base index needs more than offset?

staltz: Okay! What about https://github.com/ssb-ngi-pointer/ssb-db2/blob/82062672503062c4f16d96290c317020addeb163/compat/ebt.js#L19-L24 ?

arj03: OMG, yes we need to change that ;-)

Loading 20607 feeds
20.785s

arj03: Yes, you are correct. Had forgotten about that one.

arj03: I was wondering what if we store the actual message instead of the offset, but from a quick calculation that gets quite big quickly: 8kb * 20k ~ 160mb. Now these are worst case, but still quite a lot. I'll try and benchmark offset loading.

arj03: 1.3 seconds in comparison. So +1 to your original suggestion :-)

arj03:
Superseed by https://github.com/ssb-ngi-pointer/ssb-db2/pull/236

PR #232, opened 2021-06-24
Call post on add functions like publish does

arj03:
While testing unbermuda I noticed that something was up with EBT.

EBT on startup gets the [latest state](https://github.com/ssbc/ssb-ebt/blob/master/index.js#L81) and then maintains the state using [post](https://github.com/ssbc/ssb-ebt/blob/master/index.js#L87). It even uses this (internally)[https://github.com/ssbc/ssb-ebt/blob/master/index.js#L74] to manage state. The problem would manifest itself in EBT when having multiple EBT connections, then you would receive too many messages because the clock is not properly kept in sync. The problem could also be triggered by something other than EBT writing messages (which is how I found this problem).

DB1 called post on [both](https://github.com/ssbc/ssb-db/blob/b2256fd7e3e6e3804ca5f388b3fc351312250d05/minimal.js#L111) `add` and `publish`. 

I have tested this during an initial sync kind of scenario and the overhead does not seem to be very much. I was fearing that.

### DIFF
--- a/db.js
+++ b/db.js
@@ -159,7 +159,10 @@ exports.init = function (sbot, config) {
           state = validate.append(state, hmac_key, msg)
           if (state.error) return cb(state.error)
           const kv = state.queue[state.queue.length - 1]
-          log.add(kv.key, kv.value, cb)
+          log.add(kv.key, kv.value, (err, data) => {
+            post.set(data)
+            cb(err, data)
+          })
         } catch (ex) {
           return cb(ex)
         }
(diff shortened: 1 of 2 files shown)


staltz:
Very good that you spotted this. Makes a lot of sense that every log-appending operation should also do `post`. Did you test that the new tests fail on the previous implementation in master branch?.

staltz (approved):

arj03:
Thanks for the review. The tests on master shows that `post` is not called on `add`. 

I'll try this together with my unbermuda branch and confirm that it truly fixes the issue before merging.

PR #236, opened 2021-07-05
use validate2 and pass all tests

staltz:
cc @arj03 @mycognosist

- [x] use validate2
- [x] pass all existing tests
- [x] new base index structure
- [x] addBatch and tests
- [x] addOOOBatch and tests
- [x] use `validate2.ready()`
- [x]  make sure it supports browser, either via -wasm or ssb-validate2 (js)
- [x] OPTIONAL ~~remove ssb-validate1 entirely~~ https://github.com/ssb-ngi-pointer/ssb-validate2-rsjs/issues/31
- [x] OPTIONAL ~~Make -wasm fast~~ https://github.com/ssb-ngi-pointer/ssb-validate2-rsjs/issues/28
- [x] Debouncing batch in EBT
- [x] Test performance on mobile => reduces initial sync duration by 75% :+1:
- [x] `ssb.db.add` to do debouncing and addBatch
- [x] Fix performance of "add 1000 msgs" benchmark
- [x] Consider a license change because ssb-validate2 is AGPL-3.0?
- [x] New version of https://crates.io/crates/ssb-crypto with LGPL-3.0
- [x] New version of https://crates.io/crates/ssb-validate with LGPL-3.0
- [x] New version of https://crates.io/crates/ssb-verify-signatures with LGPL-3.0
  - [x] New version of https://crates.io/crates/ssb-legacy-msg-data with LGPL-3.0
  - [x] New version of https://crates.io/crates/ssb-multiformats with LGPL-3.0
- [x] New version of ssb-validate2-rsjs-node


db.js
```diff
@@ -147,53 +158,57 @@ exports.init = function (sbot, config) {
...
-      get(kv.key, (err, data) => {
-        if (data) cb(null, data)
-        else
-          log.add(kv.key, kv.value, (err, data) => {
-            post.set(data)
-            cb(err, data)
-          })
+    validate2.validateOOOBatch(null, [msgValue], (err) => {
+      if (err) return cb(err)
+      const kvt = validate.toKeyValueTimestamp(msgValue)
```
arj03, line 199: can use the keys second args to the cb here instead

db.js
```diff
@@ -147,53 +158,57 @@ exports.init = function (sbot, config) {
...
+          cb(null, data)
+        })
       })
-    } catch (ex) {
-      return cb(ex)
-    }
+    })
   }
 
+  // FIXME:
```
arj03, line 211: Don't bother with this :-)

Where I'm currently at is I want to remove that function and instead add a addOOOBatch (together with addBatch (https://github.com/ssb-ngi-pointer/ssb-db2/pull/228)).

arj03:
Just two small comments, overall looks good :), I'll try this out now with the extra 2 functions.

arj03:
Initial results says that this is 600ms faster than master for 2 feeds in full (3.8s vs 4.4s). That might not sound like a lot, but consider this is including getting the data over the network. This is running against an ancient ssb-server on my vps. Amazing work @mycognosist!

staltz:
Good to hear about the numbers! @arj03 noticed I used paramap. I just slapped the number 8 there but I think this could be experimented with to find the best value.

arj03:
I was using this [branch](https://github.com/ssb-ngi-pointer/ssb-db2/tree/validate2-addbatch) to test btw :)

db.js
staltz: I read your comment a couple of times but I still don't know what you mean, can you clarify? :sweat_smile:

staltz: Ooh, now I get it. :heavy_check_mark:

staltz:
@arj03 This is ready for a code review, but I wouldn't merge it yet (it's a big breaking change, we want to make sure it'll run on browsers and on mobile).

arj03:
in the browser this is 36 seconds compared to ~5 seconds in node. This is running on battery so the numbers are  a bit higher than the post above.

staltz:
@arj03 I added two commits to this PR:

* use ssb-validate2 (js) in the browser
* call validate2.ready() on start up

staltz:
## Mobile benchmarks

I'll use this comment as a wiki where I'll update the benchmark numbers for Manyverse's end-to-end replication benchmark:

- **Before** (each sub item below is an independent run of the benchmarks):
  - 342s
  - 430s
- **After**
  - 407s
  - 233s
  - 485s
  - 499s

db.js
```diff
@@ -147,91 +165,126 @@ exports.init = function (sbot, config) {
...
+    onceWhen(
+      stateFeedsReady,
+      (ready) => ready === true,
+      () => {
+        const latestMsgVal = state[msgVal.author]
+          ? state[msgVal.author].value
+          : null
+        validate2.validateSingle(hmacKey, msgVal, latestMsgVal, (err, key) => {
+          if (err) return cb(err)
+          updateState(validate.toKeyValueTimestamp(msgVal))
```
arj03, line 239: can use key here

test/basic.js
```diff
@@ -198,6 +198,27 @@ test('add 3 messages', (t) => {
...
+
+  s = validate.appendNew(s, null, rando, post4, Date.now() - 3)
+  s = validate.appendNew(s, null, rando, post5, Date.now() - 2)
+  s = validate.appendNew(s, null, rando, post6, Date.now() - 1)
+
+  const msgVals = s.queue.map((kvt) => kvt.value)
+  db.addBatch(msgVals, (err, keys) => {
+    t.error(err, 'no err')
+    t.equals(keys.length, 3)
+    t.deepEquals(keys, [undefined, undefined, undefined])
```
arj03, line 217: why are these undefined? I'm assuming we are doing the `pluck: 2` to get the kvts?

And it seems `addOOOBatch` would want `pluck: 2` as well?

staltz: Oh, well, I made addBatch return undefined on purpose because I thought that we didn't need any output. So I used `pluck: 2` to create those undefined (reminder: pluck 2 means we pick the **third** argument in the callback, which is undefined, the **first** argument would be the `err` and the **second** argument, i.e., pluck 1, would be `res`).

I can change these, I just want to know where will these keys be used as an output.

arj03: I think that would be best to keep it consistant with `add`

staltz:
I just finished this: 

> - [x] `ssb.db.add` to do debouncing and addBatch

staltz:
@arj03 The performance of "add 1000 elements" makes much more sense now, but it might still surprise that it's not as fast as before.

PR #238, opened 2021-07-12
refactor benchmarks to avoid DeferredPromise

staltz:
Two main changes here:

- Removed the DeferredPromise in benchmarks/index.js
  - I thought this was the only way of doing it, but I figured out a simpler way, which is just `await new Promise((resolve) => `
- Moved the "add 1000 messages" benchmark to the bottom
  - I noticed that since this was the first test, then it meant that all subsequent tests were running over 101k messages, not 100k messages as ssb-fixtures was configured for. I guess we should stick to 100k messages for the tests because it helps us know how much time "per message" on average is spent. We should not add messages to these tests.

### DIFF
--- a/benchmark/index.js
+++ b/benchmark/index.js
@@ -7,14 +7,13 @@ const generateFixture = require('ssb-fixtures')
 const SecretStack = require('secret-stack')
 const caps = require('ssb-caps')
 const ssbKeys = require('ssb-keys')
-const multicb = require('multicb')
 const pull = require('pull-stream')
 const asyncFilter = require('pull-async-filter')
 const validate = require('ssb-validate')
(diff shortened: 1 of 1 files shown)


arj03 (approved):
I was looking at the old file and the new and this does seem to improve readability overall. This will probably mean I need to rebase some PRs but that is okay, I want to rebase them on top of validate2 PR anyway.

staltz:
@arj03 if you want, I can do this refactor after the big merges happen.

arj03:
Might be a good way to review it :)

staltz:
I'll park this PR for now, I can copy-paste or reuse parts of it later.

PR #240, opened 2021-07-13
move some dependencies to devDependencies

staltz:
I noticed that Prettier should be a devDependency, and that some dependencies are actually not used anywhere (e.g. `ssb-sort`).


mycognosist (approved):
Nice, one less module to pull. Good catch.

PR #243, opened 2021-07-26
Validate2 review changes

arj03:
This is my review of https://github.com/ssb-ngi-pointer/ssb-db2/pull/236. I have put it into ssb-browser and did quite a few tests. Initially I had the same `asyncMap(sbot.db.add)` so that part was really slow like the tests. Not sure how to best avoid that gotcha. 

It is probably easiest to read this as the individual commits, the first two are rather simple. The next 2 is about making sure that post.set works correctly in all cases. post.set is used to "communicate" with ebt and before there was a bug where `addBatch` did not call `post.set`. Furthermore the ooo methods must never call that function, unless they are at the head of the feed (this is why I added setPost, secretstack seems to hook that object / function into oblivion). In general I'm not super happy about the way this post is used between these modules, a bit too magic. For the sake of backwards compatibility I think we should probably leave it as is.

The last commit changes our dependency to rsjs-node directly so that we don't try to compile the wasm module as it is currently not used (at least not in the browser).

### DIFF
--- a/db.js
+++ b/db.js
@@ -214,13 +214,12 @@ exports.init = function (sbot, config) {
             if (i === msgVals.length - 1) {
               // last KVT, let's update the latest state
               updateState({ key: keys[i], value: msgVals[i] })
-              log.add(keys[i], msgVals[i], (err, kvt) => {
-                post.set(kvt)
-                done()(err, kvt)
-              })
-            } else {
-              log.add(keys[i], msgVals[i], done())
             }
+
+            log.add(keys[i], msgVals[i], (err, kvt) => {
+              post.set(kvt)
+              done()(err, kvt)
+            })
           }
 
           done(cb)
@@ -481,6 +480,7 @@ exports.init = function (sbot, config) {
     publish,
     addOOO,
     addOOOBatch,
+    setPost: post.set,
     getStatus: () => status.obv,
     operators,
     post,
--- a/db.js
+++ b/db.js
@@ -211,14 +211,17 @@ exports.init = function (sbot, config) {
 
           const done = multicb({ pluck: 1 })
           for (var i = 0; i < msgVals.length; ++i) {
-            if (i === msgVals.length - 1) {
-              // last KVT, let's update the latest state
-              updateState({ key: keys[i], value: msgVals[i] })
-            }
+            const isLast = i === msgVals.length - 1
 
             log.add(keys[i], msgVals[i], (err, kvt) => {
+              if (err) return done()(err)
+
+              if (isLast) {
+                // last KVT, let's update the latest state
+                updateState({ key: keys[i], value: msgVals[i] })
+              }
               post.set(kvt)
-              done()(err, kvt)
+              done()(null, kvt)
             })
           }
 
@@ -241,9 +244,10 @@ exports.init = function (sbot, config) {
           : null
         validate2.validateSingle(hmacKey, msgVal, latestMsgVal, (err, key) => {
           if (err) return cb(err)
-          updateState({ key, value: msgVal })
           log.add(key, msgVal, (err, kvt) => {
             if (err) return cb(err)
+
+            updateState({ key, value: msgVal })
             post.set(kvt)
             cb(null, kvt)
           })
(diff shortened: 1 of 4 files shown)


db.js
```diff
@@ -481,6 +480,7 @@ exports.init = function (sbot, config) {
     publish,
     addOOO,
     addOOOBatch,
+    setPost: post.set,
```
staltz, line 483: Which module is expected to call this API? I know you said the below already, but it didn't clarify where is this function called.

> *Furthermore the ooo methods must never call that function, unless they are at the head of the feed (this is why I added setPost, secretstack seems to hook that object / function into oblivion).*

test/validate.js
```diff
@@ -93,13 +93,16 @@ test('Raw feed with unused type + ooo in batch', (t) => {
...
+  const latestPost = db.post.value
+
   const msgVals = state.queue.slice(2).map((msg) => msg.value)
   db.addOOOBatch(msgVals, (err) => {
     t.error(err, 'no err')
 
     db.addOOO(state.queue[0].value, (err, oooMsg) => {
       t.error(err, 'no err')
       t.equal(oooMsg.value.content.text, 'test1', 'text correct')
+      t.equal(db.post.value, latestPost, 'ooo methods does update post') // as that would break EBT
```
staltz, line 105: Sounds like the string should be  `'ooo methods do not update post'`?

db.js
```diff
@@ -213,8 +213,9 @@ exports.init = function (sbot, config) {
           for (var i = 0; i < msgVals.length; ++i) {
             if (i === msgVals.length - 1) {
               // last KVT, let's update the latest state
+              updateState({ key: keys[i], value: msgVals[i] })
               log.add(keys[i], msgVals[i], (err, kvt) => {
-                if (!err && kvt) updateState(kvt)
```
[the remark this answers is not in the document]
arj03: It was changed to updateState first to match publish and addImmediately. I'll fix the error checking.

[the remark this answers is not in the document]
staltz: Yeah, I see the matching, what I'm thinking is that we probably *should* in all cases call `updateState` after `log.add` completes successfully. Say there is an error writing to the log, can things go bad if the latest KVT is out of sync?

db.js
arj03: I'm using it [here](https://github.com/arj03/ssb-browser-core/blob/2a339814c0c6742a78f3a0e6cb1322a03d93c6fc/feed-replication.js#L46). The idea is to get about, get contacts and lastly latest messages. For latest messages we need to set post so that EBT are able to "continue".

db.js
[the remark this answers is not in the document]
arj03: Good question. The question is in what cases log.add would fail, like it could def. fail if you ran out of space. If you do updateState and then only if log.add works, update post, then EBT would be in a wrong state and try to get the same message that it can't add to the database because it thinks its there. I think the reason it was there before might have been to handle concurrent writes, but that is anyway fucked because validate is a async function. I'm leaning towards making it consistent with updating post & updateState at the same time.

db.js
staltz: Okay, I'm fine if we add this for now. I would appreciate some disclaimers, though, like a TODO linking to an issue in GitHub. `setPost` is a bit breaking encapsulation, because `post` used to be read-only for other modules, now it's read-write. I understand that we need to partially replicate latest messages and then "set" the state so that EBT can notice it, but we need to rethink this logic at some point (soon hopefully).

arj03: Totally agree, I'll add a warning and not add it to the API so that we know it will be changed. Then we can fix this when we do the replicate using indexes & metafeeds.

arj03:
Pushed up another commit with some fixes.

db.js
```diff
@@ -211,14 +211,17 @@ exports.init = function (sbot, config) {
...
-              updateState({ key: keys[i], value: msgVals[i] })
-            }
+            const isLast = i === msgVals.length - 1
 
             log.add(keys[i], msgVals[i], (err, kvt) => {
+              if (err) return done()(err)
+
+              if (isLast) {
+                // last KVT, let's update the latest state
+                updateState({ key: keys[i], value: msgVals[i] })
```
staltz, line 221: Since we have the kvt, we can just write `updateState(kvt)` here

staltz: Not important though, we can merge without this refactor

staltz (approved):

PR #248, opened 2021-08-16
Remove publishAs() for meta feeds

staltz:
@arj03 This PR targets `publishAs-simple` because I didn't want to push to that branch directly. Reviewing commit-per-commit is probably easiest.

This PR surprisingly is *removing* support for bendy butt in `publishAs()` but bear with me: I progressively realized that "wait, we don't need this" and then "we don't need that either", and deleted code until it got to this state.

If we put bendy butt validation in `publishAs()` (as the FIXME said) **and** implement support for bendy butt in `add()`, the two functions end up looking exactly the same (for the bendy butt part), and that's because both of them accept a `msgVal`, validate it, update state, and add it to the log.

So I removed support for bendy butt in `publishAs()`, and when you think about it, it makes sense because the bendy butt ID used is always different to `config.keys` ID, and that sounds like a use case for `add()`. I took a look in `ssb-meta-feeds`, and we could easily use `ssb.db.add(bendyButtMsgVal)` over there.

I guess this ties back to your question whether `bendyButt.encodeNew` should be inside `publishAs()`. Basically it *could*, but that would require moving a ton of code from ssb-meta-feeds into ssb-db2, and I'm not sure that's useful or easy to understand. Basically, semantically, `publish()` is always "give me **content**, and I will wrap it in a msgVal".

In the future, we could consider moving "classic content to classic msgVal" algorithm to a module `ssb-classic-feed` which would be similar to `ssb-bendy-butt` in responsibility, so that ssb-db2 doesn't need to make hard assumptions anymore.

Finally, this PR also simplified `publish()` because we can implement it in terms of `publishAs()`. The only difference between the implementation of those two turned out to be just one variable. And then also `publishAs()` internally calls `add()`.

I ran tests locally and they passed :heavy_check_mark: (CI is not configured to run for branches-of-branches), but it might be a good idea to run them yourself locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -358,23 +358,25 @@ this method unless you know exactly what you are doing.
 Delete all messages of a specific feedId. Compared to `del` this
 method is safe to use.
 
-### publish(msg, cb)
+### publish(content, cb)
 
-Convenience method for validating and adding a message to the database
-written by the feed running the secret-stack. If message contains
-recps, the message will automatically be encrypted.
+Convenience method for validating and adding a classic SSB message to
+the database written by the feed running the secret-stack. If message
+`content` contains recps, the message will automatically be encrypted.
 
-### publishAs(feedKey, msg, cb)
+### publishAs(feedKeys, content, cb)
 
-Convenience method for validating and adding a message to the database
-written by a different feed than running the secret-stack. If message
-contains recps, the message will automatically be encrypted.
+Convenience method for validating and adding a classic SSB message to
+the database written by a different feed than running the secret-stack.
+If message `content` contains recps, the message will automatically be
+encrypted.
 
-### add(msg, cb)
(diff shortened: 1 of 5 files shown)


arj03:
Ah yes of course! That is a great idea. It would be good to update the README to match these new things, like `publishAs` only takes classic and takes in content instead of msg. Same content thing for `publish`.

staltz:
Oh right, forgot about README. will do that.

arj03 (approved):
Other than that, the changes look good. I'm currently testing things on the validate-publishAs-simple-box2 branch, so it should be safe to merge from my point of view. I'll update accordingly once we have the validate2 and this merged, just wanted something I can start working from (and discover the horrors of sodium-native ;-))

PR #252, opened 2021-08-19
Support bendy butt feeds

staltz:
- [x] Supercede #241 
- [x] Code review
- [x] Dependencies that use this branch have stabilised
- [x] Undraft this PR and merge 

@arj03 This branch supercedes https://github.com/ssb-ngi-pointer/ssb-db2/pull/241, for a few reasons: 

- Rebasing was difficult, maybe I did a mistake
- Some dependencies use `publishAs-simple` and it would be dangerous to mutate the branch and make them change
- As soon as this branch is reviewed, we can delete `publishAs-simple`

I also think that this branch could be merged and released as a new version, it still supports existing use cases correctly.

PR #254, opened 2021-08-20
Make sure addBatch updates state at the correct moment

arj03:
from my review of #252


arj03:
Now the test did not catch any errors both before and after. Still think the test is good to have especially in the CI as it will better test concurrent adds.

staltz (approved):
:1st_place_medal:

PR #255, opened 2021-08-20
support SSB URIs

staltz:
- [x] Tests pass :heavy_check_mark: 
- [x] Pending on ssb-bendy-butt https://github.com/ssb-ngi-pointer/ssb-bendy-butt/pull/29
- [x] Pending on updating ssb-bendy-butt's npm version
- [x] Update `author` operator to support URIs
- [x] Pending on JITDB getting a `version` option for bitvector indexes


staltz:
This PR is ready (to go into branch `support-bendy-butt`). All tests pass (ran locally)

arj03:
This is without query author changes?

staltz:
I'm not sure what that means :eyes:

arj03:
Trying to understand how much is in :)

This will store bendy butt feeds using URIs in the database right? So a `query(where(author(` also needs to be updated to support that case, right?

staltz:
Aah, right, now I know what you mean. The `author` operator needs to get that JITDB feature of finding the prefix

staltz:
@arj03 I was working on JITDB when I realized something, so I quit that idea and did something else. Check the latest commit in this PR, I'm curious to hear what you think.

(I still need to update JITDB so that we bump the `version` encoded in the index file for authors, because the whole file needs to be rebuilt)

staltz:
I thought that versioning of indexes is a good thing for us in general because if we decide to do a "proper prefix offset" in the future, we can just update this index by bumping the version. So now I'm more comfortable with this hack, because it means we keep JITDB simple and avoid a performance cost.

staltz:
I updated jitdb and bumped the version of the author index.

operators/index.js
```diff
@@ -30,6 +30,18 @@ function type(value) {
   })
 }
 
+// We don't need the author "prefix" to be an actual prefix, it can just be any
+// predefined positions in the "information" part of the author ID.
+//
+// WARNING: when updating this, be extra careful that the resulting number isn't
+// larger than the smallest ID's length. E.g. classic feed IDs are 53 characters
+// long, and the base64 part ends at character 44, so AUTHOR_PREFIX_OFFSET must
+// be smaller than 44.
```
arj03, line 39: This is pretty clever. Nice one :)

arj03: if I were to nitpick (seeing as I can't find anything else to comment on ;-)), maybe mention here that it must be smaller than 44 - prefix bits in bytes, so by default less than 40.

arj03 (approved):
Great work on this

staltz:
Thanks for taking a look! (I fixed the code comment too)

PR #258, opened 2021-09-07
allow type() operator to be non-dedicated

staltz:
To solve issue https://github.com/ssb-ngi-pointer/ssb-meta-feeds-rpc/issues/9 in [ssb-meta-feeds-rpc](https://github.com/ssb-ngi-pointer/ssb-meta-feeds-rpc), we need to update [ssb-subset-ql](https://github.com/ssb-ngi-pointer/ssb-subset-ql) to use a non-dedicated index for `value_content_type`, and in turn that requires adding this functionality to ssb-db2.

### DIFF
--- a/operators/index.js
+++ b/operators/index.js
@@ -24,10 +24,17 @@ function key(msgId) {
   })
 }
 
-function type(value) {
-  return equal(seekType, value, {
-    indexType: 'value_content_type',
-  })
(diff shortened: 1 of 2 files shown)


mycognosist (approved):
I can't find anything that's obviously out of place.

LGTM 🟢

PR #261, opened 2021-09-14
support opts.limit in createLogStream()

staltz:
1st commit adds a failing test, 2nd commit fixes the test.

## Context

For netsim we need go-ssb, ssb-js with ssb-db, and ssb-js with ssb-db2 to behave similarly. 

## Problem

go-ssb and (js) ssb-db support the opts.limit arg in createLogStream, but ssb-db2 doesn't.

## Solution

Add opts.limit in createLogStream


mycognosist (approved):
Awesome. Thanks for adding this so promptly! The added code and test look good 🍏

PR #263, opened 2021-09-22
add addTransction() functionality

arj03:
This exposes the recently added transaction support in AAOL. I thought it would be fine that this just worked on classic messages. We can always add bendy butt if needed. Similar to addBatch.

### DIFF
--- a/db.js
+++ b/db.js
@@ -206,7 +206,19 @@ exports.init = function (sbot, config) {
       stateFeedsReady,
       (ready) => ready === true,
       () => {
-        function add(msgKeys, oooKeys) {
+        function add(err, keys) {
+          if (err) return cb(err)
+
+          const [msgKeys, oooKeys] = keys
+
+          if (msgKeys.length > 0) {
+            const lastIndex = msgKeys.length - 1
+            updateState({
+              key: msgKeys[lastIndex],
+              value: msgVals[lastIndex],
+            })
+          }
+
           log.addTransaction(
             msgKeys.concat(oooKeys),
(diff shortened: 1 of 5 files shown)


arj03:
The second commit is not strictly needed, but thought it might be good anyway

db.js
```diff
@@ -194,6 +195,62 @@ exports.init = function (sbot, config) {
...
+            (err, kvts) => {
+              if (err) return cb(err)
+
+              kvts.forEach((kvt) => post.set(kvt))
+              cb(null, kvts)
+            }
+          )
+        }
+
+        validate2.validateOOOBatch(hmacKey, oooMsgVals, (err, oooKeys) => {
```
staltz, line 222: Took me a while to understand this function, it's a bit funky with the separate keys, and the two different validations. I wonder if it would be simpler to use `multicb` to run both validateOOOBatch and validateBatch in parallel?

staltz:
> I thought it would be fine that this just worked on classic messages. We can always add bendy butt if needed.

:+1:

db.js
arj03: Yeah, that refactor found a small bug (if validateOOO fails it was not caught) ;)

staltz:
Good, it's an improvement. :)
One more thing if you don't mind, seems like `add` is only called in one place so you could just put it directly in `done((err, keys) =>` this should make it easier to read the code sequentially from top to bottom

staltz (approved):
Thanks!

arj03:
Thanks for the review. That was a good improvement :)

PR #266, opened 2021-10-06
Add debug msg for classic validation faliure

mycognosist:
Adds a `debug` message to `addImmediately()` to report when validation fails for a classic message.


arj03:
Much appreciated. This should make debugging a lot easier in the future :)

PR #273, opened 2021-10-17
multicb is a dependency, not dev dep

staltz:
It's used in db.js, and when installing ssb-db2 with other package managers like pnpm, multicb would be missing if it's a dev dep.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
     "level-codec": "^9.0.2",
     "lodash.debounce": "^4.0.8",
     "mkdirp": "^1.0.4",
+    "multicb": "1.2.2",
     "obz": "^1.0.3",
     "p-defer": "^3.0.0",
     "promisify-4loc": "1.0.0",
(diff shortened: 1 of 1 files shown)


staltz:
Apparently CI started failing across several repos (see also https://github.com/ssbc/ssb-ebt/pull/58) due to something about Rust compilation in ssb-validate2-rsjs-node

staltz:
The problem was in ssb-validate2-rsjs-node where [version 1.0.1 was released without compiling the prebuilts](https://github.com/ssb-ngi-pointer/ssb-validate2-rsjs-node/actions/runs/1328406577) because the commit needs to be `release x.y.z`, [so I made "release 1.0.2"](https://github.com/ssb-ngi-pointer/ssb-validate2-rsjs-node/actions/runs/1351557349).

staltz:
Ah, I think the prebuilds folder are not allow-listed in the npm package

arj03 (approved):

PR #275, opened 2021-11-05
Handle exceptions from box2 as err cb instead of throwing

arj03:
I was testing a client with private groups and ran into an issue where own key was not provided. This together with latest ssb-db2-box2 provides a better error and I wanted to let that error show up in cb instead of having client handling the throw. With this merged tests on ssb-db2-box2 should [work again](https://github.com/ssb-ngi-pointer/ssb-db2-box2/runs/4115715159?check_suite_focus=true). 

Note the bump of ssb-db2-box2 is not strictly needed as we just use it for benchmarking, but thought it would be good to have that the latest version anyway.

### DIFF
--- a/db.js
+++ b/db.js
@@ -409,7 +409,13 @@ exports.init = function (sbot, config) {
       stateFeedsReady,
       (ready) => ready === true,
       () => {
-        if (content.recps) content = encryptContent(content)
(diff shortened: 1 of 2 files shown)


staltz (approved):
Looks good!

PR #276, opened 2021-11-11
fix loadStateFeeds to call reEncrypt

staltz:
Fixes a false positive "feed fork" error. Full context here: https://github.com/ssbc/ssb-ebt/issues/61

1st :x: , 2nd :heavy_check_mark:

### DIFF
--- a/db.js
+++ b/db.js
@@ -119,7 +119,7 @@ exports.init = function (sbot, config) {
   }
 
   function updateState(kvt) {
-    state[kvt.value.author] = kvt
+    state[kvt.value.author] = PrivateIndex.reEncrypt(kvt)
   }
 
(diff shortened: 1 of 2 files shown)


arj03 (approved):
1 line fix, nice! :)

staltz:
Yeah, and it's obvious now in hindsight because the `state` msgs are only used for validation.

Do you mind if I also make this fix as a backport on 2.3.x? I don't want Manyverse to stay on 2.3.x forever, I want to update, but I feel like I need to make a hot fix release since this bug is kind of a showstopper.

arj03:
Sure thing. I guess a lot of people will be happy with a new release with these 2 fixes :)

PR #277, opened 2021-11-17
Fix box2 classic encryption for publishAs

arj03:
This is the problem I was experiencing with box2. Bendy butt was working fine, but not classic messages. The problem was the for classic messages we always used config.keys as the author, which doesn't work for metafeeds. So I had to change the api to make it work: https://github.com/ssb-ngi-pointer/ssb-db2-box2/commit/9c82d1909746327182511a0b8effd81de332ca3d. This PR uses the new api. With this change the tests will pass in ssb-db2-box2.

### DIFF
--- a/db.js
+++ b/db.js
@@ -315,10 +315,14 @@ exports.init = function (sbot, config) {
     )
   }
 
-  function encryptContent(content) {
+  function encryptContent(keys, content) {
     if (sbot.box2 && content.recps.every(sbot.box2.supportsBox2)) {
-      const feedState = state[config.keys.id]
-      return sbot.box2.encryptClassic(content, feedState ? feedState.key : null)
+      const feedState = state[keys.id]
+      return sbot.box2.encryptClassic(
+        keys,
+        content,
+        feedState ? feedState.key : null
+      )
     } else return ssbKeys.box(content, content.recps)
   }
 
@@ -411,7 +415,7 @@ exports.init = function (sbot, config) {
       () => {
         if (content.recps) {
           try {
-            content = encryptContent(content)
+            content = encryptContent(keys, content)
           } catch (ex) {
             return cb(ex)
           }
--- a/package.json
+++ b/package.json
@@ -63,7 +63,7 @@
     "secret-stack": "6.4.0",
     "ssb-caps": "1.1.0",
     "ssb-db": "19.3.1",
-    "ssb-db2-box2": "^0.3.2",
+    "ssb-db2-box2": "^0.5.0",
     "ssb-fixtures": "3.0.2",
     "ssb-friends": "5.1.0",
     "ssb-threads": "9.2.0",
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "async-append-only-log": "^3.1.0",
     "atomic-file-rw": "^0.2.1",
     "binary-search-bounds": "^2.0.4",
-    "bipf": "~1.5.0",
+    "bipf": "1.5.1",
     "debug": "~4.3.1",
     "fastintcompression": "0.0.4",
     "flumecodec": "0.0.1",


staltz:
Code changes look very simple and I could approve, but the tests are failing?

arj03:
Yeah I don't get it. It's in migration, running fine here locally.

arj03:
Wait a minute, I upgraded some deps locally including bipf and now they started failing here as well.

arj03:
Yep, the problem is bipf 1.5.2

staltz:
Oh wow, I thought that patch wouldn't break anything 😬

arj03:
Yeah, it is a bit odd. The changes looked simple enough. I have to go to bed now.

staltz:
I can take a look tomorrow at the bipf problem on a separate PR targeting master. Then once the bipf issue is fixed we can merge this one. Good night!

staltz:
Here's what I learned:

The new code breaks this PR:
```js
   if ((tag & TAG_MASK) !== STRING) return null
```

The old code works:

```js
   if (tag & (TAG_MASK !== STRING)) return null
```

That's in `bipf.compareString()` which is used in jitdb (`checkEqual()`) and in many indexes to check if a certain field has the exact string we expect.

arj03:
Oh my, so it seems to be depending on a bug?

staltz:
Analyzing the old code carefully, those two uppercase constants are always the same, so it's equivalent to:

```js
   if (tag & (7 !== 0)) return null
```

which is equivalent to

```js
   if (tag & true) return null
```

which is equivalent to:

- `if` condition :heavy_check_mark: (and returns null) if `tag === 0`
- `if` condition :x:  if `tag === 1`
- `if` condition :heavy_check_mark: if `tag === 2`
- `if` condition :x:  if `tag === 3`
- `if` condition :heavy_check_mark: if `tag === 4`
- `if` condition :x:  if `tag === 5`
- ...

In other words, the `if` condition passes whenever the `tag` is even. This could mean that a lot of the `compareString()` calls return `null`, which means "strings are not the same".

staltz:
I put some console.logs in bipf showing whether the new code and old code would return null, here are logs from the ssb-db2 test that is failing in master branch ("test migrate with encrypted messages"):

`type` is calculated as `tag & TAG_MASK`

```
compareString tag=32 type=0

compareString tag=5885 type=5
  old code would return null
  new code would return null
  
compareString tag=56 type=0

compareString tag=3301 type=5
  old code would return null
  new code would return null
  
compareString tag=6725 type=5
  old code would return null
  new code would return null
  
compareString tag=32 type=0

compareString tag=4013 type=5
  old code would return null
  new code would return null
  
compareString tag=32 type=0

compareString tag=3309 type=5
  old code would return null
  new code would return null

compareString tag=40 type=0
    
compareString tag=2941 type=5
  old code would return null
  new code would return null
  
compareString tag=32 type=0

compareString tag=14 type=6
  new code would return null

compareString tag=40 type=0
  
compareString tag=2941 type=5
  old code would return null
  new code would return null
  
compareString tag=40 type=0

compareString tag=3365 type=5
  old code would return null
  new code would return null
  
compareString tag=32 type=0

compareString tag=8277 type=5
  old code would return null
  new code would return null
  
compareString tag=32 type=0

compareString tag=3229 type=5
  old code would return null
  new code would return null
  
compareString tag=32 type=0

compareString tag=14 type=6
  new code would return null
```

From this we learn that while bipf is traversing an SSB message, most of the data is objects (`5`) or strings (`0`) or arrays (`4`) and sometimes boolnull (`6`). I think it's correct to say that if the `tag` is even/odd than so is the `type` even/odd because the type is just the first 3 bits of the `tag`, and the lowest bit determines whether it's even/odd.

So this means that:

- type `0` (string) is safe because in the old code it's an even number and the `if` doesn't return null and in the new code it's the correct type string
- type `5` (object) is also safe because it's an odd number and returns null as expected, and in the new code it's not type `0` (string) which is also expected
- Any even type (e.g. `4` array or `6` boolnull) is tricky because the old code does **not** return null but the new code **does**

(To be continued...)

staltz:
Okay, I'm now very close to figuring it out.

The specific part that fails is related to the `kvt.meta.private` field, which in ssb-db2 is defined in `operators.js` as:

```js
const B_TRUE = Buffer.alloc(1, 1)
function isPrivate() {
  return equal(seekPrivate, B_TRUE, { indexType: 'meta_private' })
}
```

Note the buffer B_TRUE which is just `<Buffer 01>`. But `jitdb`'s `checkEqual()` uses `bipf.compareString()` instead of `bipf.compare()` and that doesn't make sense because we are not comparing BIPF strings. The message in the log has `0e 01` which is a BIPF boolnull for `true`, and we are passing `01` as the opData.value, and in the **old** implementation `compareString()` **happened to accept** boolnulls because those are type `6` and it's an even number, so the comparison went through.

Going forwards, we should change JITDB to use `bipf.compare()` and check all use cases of `bipf.compareString()` in indexes and make sure that the values can only be strings, otherwise we should change to `bipf.compare()`.

arj03:
Wow, nice find!

staltz:
There's also a problem in JITDB that may make things a bit slower if we fix them.

JITDB equals() takes arguments such as string `'post'` and converts it's to a Buffer directly, like `<Buffer 70 6f 73 74>`, but if we're to use `bipf.compare()` than the arguments must be BIPF-encoded, and `<Buffer 70 6f 73 74>` is not BIPF encoded because it lacks the `tag` in front. This means that we would have to update all JITDB operators to do `bipf.allocAndEncode` **not** a simple `Buffer.from()`.

staltz:
I have an idea. We pin `bipf` to (or max) 1.5.1 in `ssb-db2` and `jitdb` and then we merge your PR #277. And then we open an issue in JITDB to use bipf.compare(). How about?

arj03:
Sounds good. I have pinned it now in this repo.

PR #280, opened 2021-11-22
Fix canDecrypt when reindexing decrypted messages

arj03:
Testing box2 for real now with private groups and ran into this bug.

I'll be pushing up a new test in ssb-db2-box2 that shows the problem. There is probably also a fix in jitdb needed for that test to pass.

### DIFF
--- a/indexes/private.js
+++ b/indexes/private.js
@@ -195,7 +195,7 @@ module.exports = function (dir, sbot, config) {
       if (!streaming) {
         // since we use bsb for canDecrypt we need to ensure recOffset
         // is inserted at the correct place when reindexing
-        var insertLocation = bsb.gt(canDecrypt, recOffset)
+        const insertLocation = bsb.gt(canDecrypt, recOffset)
         canDecrypt.splice(insertLocation, 0, recOffset)
       } else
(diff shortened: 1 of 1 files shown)


arj03:
Hmm, this could probably use bsb to find the index ;) One moment

arj03:
Fixed the O(n) (read ohh no ;-))

indexes/private.js
```diff
@@ -192,7 +192,14 @@ module.exports = function (dir, sbot, config) {
       const content = tryDecryptContent(ciphertext, recBuffer, pValue)
       if (!content) return record
 
-      canDecrypt.push(recOffset)
+      if (!streaming) {
+        // since we use bsb for canDecrypt we need to ensure recOffset
+        // is inserted at the correct place when reindexing
+        var insertLocation = bsb.gt(canDecrypt, recOffset)
```
staltz, line 198:
```suggestion
        var insertLocation = bsb.gt(canDecrypt, recOffset)
```

staltz (approved):
Looks fine! Just a simple suggestion for a `var` change, but that's it. Good job on spotting the bsb

arj03:
Thanks :)

PR #281, opened 2021-11-23
show where keys come from

nichoth:
In the readme example. I think this is correct. Is it?


staltz (approved):
That's one way of getting the keys, yes. Thanks

PR #287, opened 2021-12-06
Work with latest jitdb bipf values change

arj03:
These are the changes needed to make this module work with the bipf changes in jitdb. Note this also unpins bipf. The changes are almost trivial and is a good way I think to show that the jitdb PR is a sound approach.

This is still a draft PR as it depends on the jitdb PR: https://github.com/ssb-ngi-pointer/jitdb/pull/182

### DIFF
--- a/operators/index.js
+++ b/operators/index.js
@@ -153,14 +153,13 @@ function authorIsBendyButtV1() {
 }
 
 function isRoot() {
-  return equal(seekRoot, null, {
+  return equal(seekRoot, undefined, {
     indexType: 'value_content_root',
   })
 }
 
-const B_TRUE = Buffer.alloc(1, 1)
 function isPrivate() {
-  return equal(seekPrivate, B_TRUE, { indexType: 'meta_private' })
+  return equal(seekPrivate, true, { indexType: 'meta_private' })
 }
 
 function isPublic() {
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "async-append-only-log": "^3.1.0",
     "atomic-file-rw": "^0.2.1",
     "binary-search-bounds": "^2.0.4",
-    "bipf": "1.5.1",
+    "bipf": "^1.5.4",
     "debug": "~4.3.1",
     "fastintcompression": "0.0.4",
     "flumecodec": "0.0.1",
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "flumecodec": "0.0.1",
     "flumelog-offset": "3.4.4",
     "hoox": "0.0.1",
-    "jitdb": "^3.5.0",
+    "jitdb": "^4.0.0",
     "level": "^6.0.1",
     "level-codec": "^9.0.2",
     "lodash.debounce": "^4.0.8",


operators/index.js
```diff
@@ -153,14 +153,13 @@ function authorIsBendyButtV1() {
 }
 
 function isRoot() {
-  return equal(seekRoot, null, {
+  return equal(seekRoot, undefined, {
```
arj03, line 156: This is the correct way to check for this. In the old jitdb code null worked the same way as undefined, so this means the root index will continue to work and don't need to be reindexed, but we need to do this properly now, hence this change.

operators/index.js
```diff
@@ -153,14 +153,13 @@ function authorIsBendyButtV1() {
...
     indexType: 'value_content_root',
   })
 }
 
-const B_TRUE = Buffer.alloc(1, 1)
 function isPrivate() {
-  return equal(seekPrivate, B_TRUE, { indexType: 'meta_private' })
+  return equal(seekPrivate, true, { indexType: 'meta_private' })
```
staltz, line 162: Nice that this code becomes even more dev-friendly now

operators/index.js
staltz: @arj03 Do you think that if this were changed to use `absent()`, would it cause a reindex?

arj03: It should not cause a reindex. What I'd like to do before we merge this is to run some queries using the new absent and check if this behaves exactly like it did before. I belief right now is that this is the same as before the change, but I need to double check. I'll let you know later this afternoon :)

staltz: I'm thinking that maybe it would cause a new index to be built?
Previously there is `value_content_root_.index` and then there would be `value_content_root__absent.index` unless we add a way to override the indexName.

arj03: So I can some local tests on a around 1GB test database I have migrated from db1 to db2. It is maybe ½ a year old. I have been running these 2 queries:

```
ssb.db.query(
  where(
    and(
      isRoot(),
      author('@6CAxOI3f+LUOVrbAl0IemqiS7ATpQvr9Mdw9LC4+Uv0=.ed25519')
    ),
```

and

```
ssb.db.query(
  where(
    and(
      isRoot(),
      type('post')
    ),
```
on both this branch and master I get the exact same number of messages for both queries.

I have tried replacing isRoot() with this function:

```
function isAbsent() {
  return absent(seekRoot, {
    indexType: 'value_content_root',
  })
}
```
and that also gives the exactly same number of messages.

You are correct that changing isRoot to be absent would require a reindex. I think we should leave it as is in order not to force unnecessary reindex, but I'm not super strict on that requirement if you'd rather have absent :) 

I noticed this seems a bit faster compared to the benchmarks in #277 but I think that is just the CI running on a different machine. I'll double check locally. JITDB at least is exactly the same speed locally.

arj03: Yep, benchmarks locally are within random fluctuations of neurons hitting the earth

staltz: Yeah, I'm with you that we don't need to change `isRoot` to use `absent()`. Anyway where we are at the moment with ssb-db2 there are several queries that would give the same results, so this is one of those cases. The refactor to `absent()` would be aesthetically neat but not worth it.

staltz (approved):
Let's go!

arj03:
Whoop. What are your thoughts on the version number @staltz ?

staltz:
@arj03 Good question. Technically the breaking changes in jitdb `equal()` carry over to ssb-db2 too, but on the other hand, the public API for ssb-db2 has never documented `equal()` directly. The usage of ssb-db2 has been documented in terms of higher-level operators like `isPublic`, `isRoot`, `author`, etc, and the fact that you can import jitdb `equal()` is just a nice convenience (and I think no one is doing that in any module), so I would say **bump patch** (since no features were added, so not a minor bump).

PR #290, opened 2021-12-09
Add links to several secret-stack plugins that works well with db2

arj03:
Should make it easier to get started using ssb-db2 for projects


staltz (approved):
Awesome!

PR #293, opened 2021-12-16
(and in darkness) bind write to make sure `this` works

### LINKED ISSUE #292: Too hot binding
I was testing too hot in async to see if I can see anything, I noticed that there appears to be a missing bind on this line: https://github.com/ssb-ngi-pointer/ssb-db2/blob/62e80f09d5e00d2039f8b1aec6a6005e618e0317/log.js#L83 It should be `o.write.bind(o)`, right?

arj03:
To fix #292. My understanding is that this should not have any performance overhead compared to earlier versions.


staltz (approved):
What a great PR title 😂

Issue #291, opened 2021-12-13
Broken seq.index, keys index, and duplicate records on log


staltz:
This issue has the same symptoms as https://github.com/ssb-ngi-pointer/jitdb/issues/161, and seems to be happening only on iOS, I haven't gotten bug reports for Android yet. 

I got the `db2` folder from Dave who has this problem and ran some scripts to study the data. Mostly, I'm concerned with the `keys` leveldb index, because when retrieving a message by a certain key, we get the wrong key.

For all records on the log, I printed out:

```
seq | offset | key in the keys leveldb index | key in the msg, if different
0 | 0 | %cwv/eckJoOfVqGx7bABbQduZBFrHZly0PfeE7nrwotc=.sha256
1 | 526 | %LqzwwHg3Lt+uU4S3DXdE8jkBfvma/1Yin1mrDzPF4V8=.sha256
2 | 995 | %V3/jZdpj1CsyVI3QH/rajkDahH1R30u1vf05OUA8h8U=.sha256
3 | 1464 | %Ctc7zcaT0N4Myvm3/+VEPItu1/1x0SJHysrCoJQzmUA=.sha256
4 | 2297 | %GNNx+fj9DmdMlazwqh439K+j0G1ovAGr4stelJj4EkM=.sha256
5 | 2766 | %PjwtwmxcVOyAZ399Z1Qy4og+z+sl8/P+Lfq3D/zV77w=.sha256
6 | 3250 | %p3aECWInrsJZdSO7Sp7CQR3a29MpIuVrAxwXG2tSTl0=.sha256
7 | 3736 | %m/OBdukgm7bDwQaT1P2aOHFXoLo5yzW7bfqy7JfrZeE=.sha256
...
1380993 | 992019409 | %bXb5tFQpLYK0kFOiOqhZRoJZoFsTBRLCWbCpIWNfUvY=.sha256
1380994 | 992019902 | %Fe9zF4gDp0a0EgvCSsaWmTgg1eHMKh5mKeDBnFr57LQ=.sha256
1380995 | 992020386 | %gp8jQXiLBKrUMuFWFjizZDUAD6FdSC694JbvPk8og2I=.sha256
1380996 | 992020870 | %7O1RVgAC1DGTieWj3dPY6ECyweQNV3XGsP0d3fvO7sM=.sha256
[...]

PR #295, opened 2021-12-21
Make sure processedOffset is only updated together with processedSeq

arj03:
From https://github.com/ssb-ngi-pointer/ssb-db2/issues/291#issuecomment-998773451. I believe this should fix off-by-x bugs in level indexes. 

From that comment:

Look at this [line](https://github.com/ssb-ngi-pointer/ssb-db2/blob/86d502f0846c3c234f5f6ff47be3f5e4bbe58c20/indexes/plugin.js#L88) and imagine we have added a new plugin. In this case we will reindex, but most plugins will just [skip](https://github.com/ssb-ngi-pointer/ssb-db2/blob/86d502f0846c3c234f5f6ff47be3f5e4bbe58c20/indexes/plugin.js#L83). BUT they will update their `processedOffset` thus *lovering* it compared to `processedSeq`. If you then kill the app while some of them have updated, but *NOT* keys or base then there will be a mismatch!

### DIFF
--- a/indexes/plugin.js
+++ b/indexes/plugin.js
@@ -96,15 +96,13 @@ module.exports = class Plugin {
 
       if (status && status.version === version) {
         processedSeq = status.processed
+        processedOffset = status.offset
+        this.offset.set(status.offset)
         if (this.onLoaded) {
           this.onLoaded(() => {
-            processedOffset = status.offset
-            this.offset.set(status.offset)
             this._stateLoaded.resolve()
           })
         } else {
-          processedOffset = status.offset
-          this.offset.set(status.offset)
           this._stateLoaded.resolve()
         }
       } else {
--- a/indexes/plugin.js
+++ b/indexes/plugin.js
@@ -40,7 +40,7 @@ module.exports = class Plugin {
     this.batch = []
 
     this.flush = (cb) => {
-      if (processedOffset < 0 || this.level.isClosed()) return cb()
+      if (processedOffset === this.offset.value || this.level.isClosed()) return cb()
       if (!this.onFlush) this.onFlush = (cb2) => cb2()
 
       const processedOffsetAtFlush = processedOffset
@@ -80,7 +80,7 @@ module.exports = class Plugin {
 
     this.onRecord = function onRecord(record, isLive) {
       let changes = 0
-      if (record.offset > this.offset.value) {
+      if (record.offset > processedOffset) {
         if (record.value) this.processRecord(record, processedSeq)
         changes = this.batch.length
         processedSeq++
@@ -98,15 +98,18 @@ module.exports = class Plugin {
         processedSeq = status.processed
         if (this.onLoaded) {
           this.onLoaded(() => {
+            processedOffset = status.offset
             this.offset.set(status.offset)
             this._stateLoaded.resolve()
           })
         } else {
+          processedOffset = status.offset
           this.offset.set(status.offset)
           this._stateLoaded.resolve()
         }
       } else {
         this.level.clear(() => {
+          processedOffset = -1
           this.offset.set(-1)
           this._stateLoaded.resolve()
         })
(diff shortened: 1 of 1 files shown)


arj03:
Please double and triple check this :)

indexes/plugin.js
```diff
@@ -84,8 +84,8 @@ module.exports = class Plugin {
         if (record.value) this.processRecord(record, processedSeq)
```
staltz, line 84: The `if` on line 83: I understand that `this.offset.value` means "the `offset` that is currently persisted in the META section on leveldb". Would it be safer to use `record.offset > processedOffset`? Notice that `processedOffset` is also loaded from the leveldb, but is guaranteed to be a bit more ahead of `this.offset.value`.

staltz:
Open question: how do we handle reindexing? Do I just deploy a reindex of all leveldb indexes in Manyverse? Could we instead do that in ssb-db2 by bumping the version somehow? On the other hand, `version` is specific to each individual leveldb plugin, so we would have to bump all of them and some of them aren't even in this module. It would be good (maybe? what do you think?) if there existed a "metaVersion" in the META section that catches all leveldb plugins.

indexes/plugin.js
arj03: Correct, `processedOffset` can be ahead. It should not be a problem since we only have 1 place calling this function: https://github.com/ssb-ngi-pointer/ssb-db2/blob/86d502f0846c3c234f5f6ff47be3f5e4bbe58c20/db.js#L502. Hmm, I'm a bit torn here if we fix that or not.

arj03:
> Open question: how do we handle reindexing? Do I just deploy a reindex of all leveldb indexes in Manyverse? 

I think that would be the safest

> Could we instead do that in ssb-db2 by bumping the version somehow?

I think a general level version would be good for this yes. As a separate PR for sure :)

indexes/plugin.js
arj03: I think you are right, lets fix this properly while we are at it. Before if for some reason the same offset was sent there would be a bug where seq would be incorrect. I've pushed another commit.

indexes/plugin.js
```diff
@@ -40,7 +40,7 @@ module.exports = class Plugin {
     this.batch = []
 
     this.flush = (cb) => {
-      if (processedOffset < 0 || this.level.isClosed()) return cb()
+      if (processedOffset === this.offset.value || this.level.isClosed()) return cb()
```
arj03, line 43: note this was also a bit wierd before because `processedOffset` might be set, but we didn't actually change anything.

indexes/plugin.js
```diff
@@ -98,15 +98,18 @@ module.exports = class Plugin {
         processedSeq = status.processed
         if (this.onLoaded) {
           this.onLoaded(() => {
+            processedOffset = status.offset
```
staltz, line 101: What?? This wasn't there before? OMG how could I have not seen that before, I looked at this file so many times and at `processedOffset` so many times. :see_no_evil:

staltz: By the way, I think this code should be changed so that we first set `processedOffset` and `this.offset` **and then** we call `this.onLoaded()`. That's because `this.onLoaded()` could do some inversion of control. It could trigger other computation to happen which *could* trigger `onRecord()` to be called, and by the time `onRecord` is called we MUST have `processedOffset` set correctly.

Seems to me like as soon as we load META from leveldb, there is no reason to delay the setting of `processedOffset`, we should just write it ASAP.

But still, call `this._stateLoaded.resolve()` only after `this.onLoaded()` is done, to preserve current code behavior.

indexes/plugin.js
staltz: How about `if (processedOffset < 0 || processedOffset === this.offset.value || this.level.isClosed())` ? Because the initial value of `this.offset.value` is not -1 if I remember correctly.

arj03: It should be -1 if the file does not exist so that extra check should not be needed I think

indexes/plugin.js
arj03: makes sense, good catch

staltz (approved):
Let's gooooo

PR #297, opened 2021-12-23
Document onDrain

### LINKED ISSUE #296: documentation request -- `onDrain`
There is [an example](https://github.com/ssb-ngi-pointer/ssb-db2#about-self) for `about-self` that shows

```js
sbot.db.onDrain('aboutSelf', () => {
  const profile = sbot.db.getIndex('aboutSelf').getProfile(alice.id)
  console.log('Alice has name:' + profile.name)
})
```

But I don't see any other mention of the `onDrain` method, and it seems important since it is the canonical way to access the 'aboutSelf' index.

arj03:
This should fix #296. @nichoth

### DIFF
--- a/README.md
+++ b/README.md
@@ -448,6 +448,15 @@ at runtime and that changes what messages can be decrypted. Calling
 this function is needed after adding a new key. The function can be
 called multiple times safely.
 
+### onDrain(indexName?, cb)
+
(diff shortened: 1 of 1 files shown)


staltz (approved):
# :chipmunk:

PR #298, opened 2021-12-23
protect privateIndex.decrypt() from deleted records

staltz:


arj03:
Seems correct. Would be nice with a test, if for nothing else than make it easy to check how this behaved before. I was wondering if you had a crash in bsb.eq( or bipf.seek before.

staltz:
It was in bipf.seekKey. Let me see if I can make a simple test

staltz:
Updated PR with 1st test :x: 2nd fix :heavy_check_mark:

arj03:
Thanks! Great find :)

arj03 (approved):

PR #303, opened 2022-01-17
add operators to the readme

nichoth:
This documents all operators added here for jitDB. It makes it easier to find them, but the drawback is that if the list of operators changes, there is one more place that needs to be changed (the documents).


arj03:
This is great. Thank you! 

Can you fix the db2 reference to point to this repo instead of planetary? And move the references to the bottom of the file like other reference.

nichoth:
Sorry my copy pasting got mixed up. it should have the right link now

arj03:
almost, the links should be at the bottom, like [here](https://github.com/ssb-ngi-pointer/ssb-db2/blob/7f23695f8bd59fc8f3ee3f476e1b0f5da69ec6f9/README.md?plain=1#L525) and remove the extra whitespace :)

nichoth:
Ok I think this is the approved format? I added links to the bottom of the page, not sure if that's what you meant

arj03:
Check #304

Issue #302, opened 2022-01-17
Database streaming glitches (+43 skipping)


staltz:
Similar to #291, I'll report here my findings.

So far I know that the versions are:

- Manyverse `0.2201.14-beta.b`
  - Desktop (Linux) with a totally new identity, not migrated from flume
  - iOS some days after rebuilding indexes on a recovered feed 
- ssb-db2 `2.8.5`
- jitdb `4.0.1`
- async-append-only-log `3.1.1`
- bipf `1.5.5`

And the symptom is: open a thread and it will actually display a raw message for something completely unrelated, indicating a problem with the leveldb `keys` index.

staltz:
# Bad indexes

On cryptix's corrupted database, I found that (output from running [study-keys-leveldb.js](https://github.com/staltz/ssb-db2-issue-291)):

```
Loaded seq.index with:
  Entries: 1298604
  Offset: 972042648

Loaded keys leveldb with:
  Version: 1
  Entries: 1298560
  Offset: 972042648

minBrokenSeq = 291883
minBrokenOffset = 210340373


seq    | offset    | key in the keys leveldb index                        | key in the msg, if different
0      | 0         | %bQnYTsq0p+A8JppKPgze+ZyZxAOPVgMwleEXZKwOP7o=.sha256
1      | 528       | %7d2LLAa0hhSFwlpHjulTm9IFtin95UkLUzV43yajj4k=.sha256
2      | 994       | %ofF5MfiaCnHBqecGIDpojT8Hd15wbAuGzmhNxrJ6beo=.sha256
3      | 1476      | %OJVc0qShJtS1o0A9pO5LZp4OmwIhyUkUmz6FV2Nthz8=.sha256
...
291880 | 210338468 | %H3rlUZ1gJNmwEQ26nuEfWj0Gwt3jL8f5itcH3wGLuRU=.sha256
291881 | 210338937 | %9QtgDc30RPPVm9pR6CHebtXD98/Bqabj3Mpbl7GmTPM=.sha256
291882 | 210339437 | %ckZr3Hro61xRR9qgaQsmE03ZOmdXQoF4ajQzHhbRrTs=.sha256
291883 | 210340373 | %73n/YRJxKIZ8gcxPdoO2lmGTveUAx0twEBcE9T+C3zM=.sha256 | %9obExCOb1r6LGmYw9EEK/EUgBqZ2GgQOvvfAYbR+5x8=.sha256
[...]

PR #305, opened 2022-01-18
Test that index upgrade works

arj03:
This PR should fix this [issue](https://github.com/ssb-ngi-pointer/ssb-db2/issues/302#issuecomment-1014674314). I could not find a test for the version upgrade functionality so wrote a test for that.


staltz (approved):
Looks good, thanks!

PR #308, opened 2022-01-26
when db is closed, dont call onDrain callbacks

staltz:
## Context

When Manyverse is closed, I notice an error from here: https://github.com/staltz/manyverse/blob/332deeec5dca6166dcb39207e7581190ed1d5367/src/backend/plugins/connUtils.ts#L38

This error is silent and there are no problems for users, but I figured it would be good to fix that error anyway.

## Problem

If `onDrain` is requested after the sbot (and ssb-db2) has been closed, we still seem to call the `cb`, even though the `cb` cannot and should not make use of any ssb-db2 APIs because the log is closed, indexes are closed, etc.

## Solution

Add a boolean variable `closed` and check that it's false whenever calling the `cb`.

1st :x:, 2nd :heavy_check_mark:

### DIFF
--- a/db.js
+++ b/db.js
@@ -63,6 +63,7 @@ exports.manifest = {
 
 exports.init = function (sbot, config) {
   let self
+  let closed = false
   config = config || {}
   config.db2 = config.db2 || {}
   const indexes = {}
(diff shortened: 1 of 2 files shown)


staltz:
Lol, tests are supposed to fail (they consistently fail for me locally), y u pass.

arj03 (approved):
Great :)

Issue #229, opened 2021-06-15
support opts.temp


staltz:
See https://github.com/ssbc/ssb-db/blob/b74c9c7d312ab247781d8237f8436668dde03af8/index.js#L54-L60

PR #312, opened 2022-02-25
add "temp" as an alternative to config.path

staltz:
Writing tests in other libraries was getting tedious, so I copied functionality from https://github.com/ssbc/ssb-db/blob/4c8a8b503df80e407e4cfb0167633f39a4a3377b/index.js#L56-L60

Fixes issue #229


arj03 (approved):
This is a great addition. Thanks

PR #313, opened 2022-03-02
use clarify-error on all public APIs

staltz:


db.js
```diff
@@ -365,16 +361,11 @@ exports.init = function (sbot, config) {
         } else if (SSBURI.isBendyButtV1FeedSSBURI(msgVal.author)) {
           const previous = (state[msgVal.author] || { value: null }).value
           const err = bendyButt.validateSingle(msgVal, previous, hmacKey)
-          if (err) {
-            debug(
-              `validation failed for bendy butt message in addImmediately(): ${err.message}`
-            )
-            return cb(err)
-          }
```
staltz, line 373: With the new changes, I was getting a bunch of lines formatted like this under `prettier`, and it was getting hard to follow the "normal" (non-error) control flow, so I decided to turn off prettier for these lines that handle errors with clarify-error

compat/history-stream.js
```diff
@@ -62,11 +63,13 @@ exports.init = function (sbot, config) {
             sbot.db
               .getJITDB()
               .paginate(query, 0, limit, false, false, (err, answer) => {
-                cb(err, pull.values(answer.results.map(formatMsg)))
+                if (err) cb(clarify(err, 'ssb-db2 createHistoryStream failed')) // prettier-ignore
```
arj03, line 66: This one is exposed over the network. Can that cause trouble that we might leak information about the local system? I checked compat/ebt and that is fine as getvectorclock is only used internally.

db.js
arj03: I like the shorter lines here. And I guess not doing debug is fine, as we'll catch these errors further up the stack.

arj03 (approved):
I look forward to seeing errors (ohh that sounds so wrong :-P) after all of these modules have been upgraded to clarify error

PR #314, opened 2022-03-07
replace EventEmitter with pull-notify

staltz:
## Problem

Using EventEmitter for migrate "progress" events is heavy, see: https://github.com/ssb-ngi-pointer/ssb-db2/pull/201

## Solution

There is no real need for the stream of progress events to be on the `sbot` as a stringly-named event. We can use pull-notify instead, which should be more straight to the point and hopefully a bit lighter.

**Note** this is a breaking change and will cause ssb-db2 version 3.0.0 to be published.

### DIFF
--- a/benchmark/index.js
+++ b/benchmark/index.js
@@ -13,9 +13,7 @@ const caps = require('ssb-caps')
 const ssbKeys = require('ssb-keys')
 const multicb = require('multicb')
 const pull = require('pull-stream')
-const asyncFilter = require('pull-async-filter')
 const validate = require('ssb-validate')
-const fromEvent = require('pull-stream-util/from-event')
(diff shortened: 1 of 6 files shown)


benchmark/index.js
```diff
@@ -13,9 +13,7 @@ const caps = require('ssb-caps')
 const ssbKeys = require('ssb-keys')
 const multicb = require('multicb')
 const pull = require('pull-stream')
-const asyncFilter = require('pull-async-filter')
```
arj03, line 16: it seems like we can remove this dep now as well?

staltz: Will do!

arj03 (approved):
Nice :)

PR #315, opened 2022-03-14
Support sortBy in jitdb

arj03:
This is to support the breaking change in jitdb 5.0

### DIFF
--- a/compat/history-stream.js
+++ b/compat/history-stream.js
@@ -61,12 +61,12 @@ exports.init = function (sbot, config) {
           if (limit) {
             sbot.db
               .getJITDB()
-              .paginate(query, 0, limit, false, false, (err, answer) => {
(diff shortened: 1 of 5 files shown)


staltz (approved):
All good here, but could we make jitdb@6.0.0 with the flumelog error thing updated, and then use that here?

I can make jitdb@6.0.0 if you wish

arj03:
Sure. Go ahead and create jitdb 6 and then update this PR :)

arj03:
I think we can removed that seeing as the only properly tested log is AAOL.

staltz:
@arj03 Final check? I'll let you press merge

arj03:
Great, I can do a release a bit later. 3.1 or 4?

staltz:
New minor, there isn't any breaking change that I can think of

PR #316, opened 2022-03-16
adding support for indexing publicWebHosting on about messages

rabble:
In order to run a viewer which respects publicWebHosting we need to have that field indexed in the about messages. This PR is pretty simple. If there is a publicWebHosting in the content it records the value, including if it's false. If it's not defined, it doesn't record it. 

This is needed so that https://github.com/planetary-social/planetary-pub can correctly not serve content where users have chosen not to have their content visible in a web viewer. 

This is my first PR for ssb-db2, so any feedback is welcome. I found some oddness with the order of the tests in about-self tests which were beyond me. These tests should cover the functionality added.

### DIFF
--- a/indexes/about-self.js
+++ b/indexes/about-self.js
@@ -82,6 +82,8 @@ module.exports = class AboutSelf extends Plugin {
       profile.image = content.image.link
     else if (typeof content.image === 'string') profile.image = content.image
 
+    if (content.publicWebHosting !== 'undefined')  profile.publicWebHosting = content.publicWebHosting
+    
     this.profiles[author] = profile
   }
 
--- a/test/about-self.js
+++ b/test/about-self.js
@@ -27,6 +27,39 @@ let sbot = SecretStack({ appKey: caps.shs })
   })
 const db = sbot.db
 
+test('set publicWebHosting to false', (t) => {
+  const about = { type: 'about', about: sbot.id, publicWebHosting: false}
+  
+  db.publish(about, (err, postMsg) => {
+    t.error(err, 'no err')
+
+    sbot.db.onDrain('aboutSelf', () => {
+      const profile = sbot.db.getIndex('aboutSelf').getProfile(sbot.id)
+      t.equal(profile.publicWebHosting, about.publicWebHosting) // should be false
+
+      t.end()
+    })
+  })
+})
+
+
+test('set publicWebHosting to true', (t) => {
+  const about = { type: 'about', about: sbot.id, publicWebHosting: true}
+  
+  db.publish(about, (err, postMsg) => {
+    t.error(err, 'no err')
+
+    sbot.db.onDrain('aboutSelf', () => {
+      const profile = sbot.db.getIndex('aboutSelf').getProfile(sbot.id)
+      t.equal(profile.publicWebHosting, about.publicWebHosting) // should be true
+
+      t.end()
+    })
+  })
+})
+
+
+
 test('get self assigned', (t) => {
   const about = { type: 'about', about: sbot.id, name: 'arj', image: '%blob' }
   const aboutOther = { type: 'about', about: '@other', name: 'staltz' }
@@ -102,6 +135,7 @@ test('get live profile', (t) => {
   })
 })
 
+
 test('should load about-self from disk', (t) => {
   sbot.close((err) => {
     t.error(err)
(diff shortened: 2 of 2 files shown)


staltz:
Thanks @rabble , this is a relevant PR. I agree with the addition.

Just have a few simple requests about putting the code in the correct shape.

indexes/about-self.js
```diff
@@ -82,6 +82,8 @@ module.exports = class AboutSelf extends Plugin {
       profile.image = content.image.link
     else if (typeof content.image === 'string') profile.image = content.image
 
+    if (content.publicWebHosting !== 'undefined')  profile.publicWebHosting = content.publicWebHosting
+
```
staltz, line 86: This would require existing index entries to be updated, so we have to bump up the index's "version". You do that by changing the number `3` to `4` on line 20.

test/about-self.js
```diff
@@ -27,6 +27,39 @@ let sbot = SecretStack({ appKey: caps.shs })
...
 
+test('set publicWebHosting to false', (t) => {
+  const about = { type: 'about', about: sbot.id, publicWebHosting: false}
+  
+  db.publish(about, (err, postMsg) => {
+    t.error(err, 'no err')
+
+    sbot.db.onDrain('aboutSelf', () => {
+      const profile = sbot.db.getIndex('aboutSelf').getProfile(sbot.id)
+      t.equal(profile.publicWebHosting, about.publicWebHosting) // should be false
```
staltz, line 38:
```suggestion
      t.false(profile.publicWebHosting)
```

test/about-self.js
```diff
@@ -27,6 +27,39 @@ let sbot = SecretStack({ appKey: caps.shs })
...
+
+test('set publicWebHosting to true', (t) => {
+  const about = { type: 'about', about: sbot.id, publicWebHosting: true}
+  
+  db.publish(about, (err, postMsg) => {
+    t.error(err, 'no err')
+
+    sbot.db.onDrain('aboutSelf', () => {
+      const profile = sbot.db.getIndex('aboutSelf').getProfile(sbot.id)
+      t.equal(profile.publicWebHosting, about.publicWebHosting) // should be true
```
staltz, line 54:
```suggestion
      t.true(profile.publicWebHosting)
```

test/about-self.js
```diff
@@ -122,6 +156,8 @@ test('should load about-self from disk', (t) => {
   })
 })
 
+
+
```
staltz, line 160: Can you run `npm run format-code` before making a git commit? It should have run automatically as a git hook, but for some reason on your local dev env it didn't. That would get rid of these added empty lines.

staltz (changes requested):

rabble:
Thanks for the feedback @staltz, super helpful. The format-code tool is nifty, it did change a few other files which i guess also had some formatting needs.

test/scratch.file
```diff
@@ -0,0 +1,67 @@
...
+    sbot.db.onDrain('aboutSelf', () => {
+      const profile = sbot.db.getIndex('aboutSelf').getProfile("@alice")
+      t.equal(profile.name, newAbout.name)
+      t.equal(profile.image, newAbout.image.link)
+      t.equal(profile.publicWebHosting, newAbout.publicWebHosting)
+
+      t.end()
+    })
+  })
+})
```
staltz, line 67: @rabble What's this file? Can we delete it?

indexes/about-self.js
staltz: @rabble we still need the index version bumped.

staltz (changes requested):

test/scratch.file
rabble: oh yeah, this should go away, sorry

rabble:
Thanks for your patience... i think i've got it sorted now. Can you take another look @staltz

operators/index.js
```diff
@@ -64,7 +64,7 @@ function author(value, opts = { dedicated: false }) {
       prefix: 32,
       prefixOffset: AUTHOR_PREFIX_OFFSET,
       indexType: 'value_author',
-      version: 2,
+      version: 3,
```
staltz, line 67: @rabble Not this one. I had mentioned specifically line 20 on indexes/about-self.js https://github.com/ssb-ngi-pointer/ssb-db2/pull/316/files#r827745805 Could you revert this one? Otherwise we would cause the value_author index to be rebuilt in vain.

rabble:
This was fixed in #388 and is no longer needed.

PR #318, opened 2022-03-25
Add startDecryptBox1 config

arj03:
# Background

Decrypting box1 messages is quite heavy (I ran a local query here on a 1GB file, querying for all messages from a particular ID is 41 sec with decryption and 30 sec without). Especially for new users where out of roughly a million or so messages, more than 10% will be encrypted. For a new user joining the network they will never be able to decrypt these messages, so it would make sense to allow one to only start a certain time. Like from your first message. That was my first solution, but I realized that figuring out when the first message is from can be problematic. You might be reinstalling from a backup keys, or even if you have a database, just figuring out when the first message is from without building the indexing.

Instead I thought we could at least have a simple timestamp that one can provide that will tell the system from which point on it should start decrypting. The immediate use case for this would be a client where the user has generated a new key. We then know the time and can provide that to db2. Another use case would be something like a pub where you don't want to decrypting messages, so you set the time to 2100-01-01 or something.

### DIFF
--- a/README.md
+++ b/README.md
@@ -493,6 +493,12 @@ const config = {
     dangerouslyKillFlumeWhenMigrated: false,
 
     /**
+     * Only try to decrypt box1 messages created before this date
+     * Default: null
+     */
+    startDecryptBox1: "2022-03-25",
+
+    /**
      * A debouncing interval (measured in milliseconds) to control how often
      * should messages given to `sbot.add` be flushed in batches.
      * Default: 250
--- a/README.md
+++ b/README.md
@@ -493,7 +493,7 @@ const config = {
     dangerouslyKillFlumeWhenMigrated: false,
 
     /**
-     * Only try to decrypt box1 messages created before this date
+     * Only try to decrypt box1 messages created after this date
      * Default: null
      */
     startDecryptBox1: "2022-03-25",
--- a/indexes/private.js
+++ b/indexes/private.js
@@ -23,6 +23,10 @@ module.exports = function (dir, sbot, config) {
   let encrypted = []
   let canDecrypt = []
 
+  let startDecryptBox1 = null
+  if (config.db2.startDecryptBox1)
+    startDecryptBox1 = new Date(config.db2.startDecryptBox1)
+
   const debug = Debug('ssb:db2:private')
 
   const encryptedFile = path.join(indexesPath(dir), 'encrypted.index')
@@ -134,6 +138,7 @@ module.exports = function (dir, sbot, config) {
   const B_CONTENT = Buffer.from('content')
   const B_AUTHOR = Buffer.from('author')
   const B_PREVIOUS = Buffer.from('previous')
+  const B_TIMESTAMP = Buffer.from('timestamp')
 
   function decryptBox1(ciphertext, keys) {
     return ssbKeys.unbox(ciphertext, keys)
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -492,6 +492,12 @@ const config = {
      */
     dangerouslyKillFlumeWhenMigrated: false,
 
+    /**
+     * Only try to decrypt box1 messages created before this date
+     * Default: null
+     */
+    startDecryptBox1: "2022-03-25",
```
arj03, line 499: I settled for string representation because passing in a date doesn't seem to work, it gets converted into an empty object.

arj03:
Here you can see the difference:


```
unbox 1000 private box1 elements first run | 165ms
unbox 1000 private box1 elements second run | 133ms
query 1000 elements first run | 38ms
query 1000 elements second run | 26ms
```

staltz:
I haven't check the code yet but the PR description looks good!

mixmix:
feature idea.
build The Great Tangle to determine when you should start decrypting or who you should be trying. 

basically when a log starts up and starts referencing other messages and feeds that provides a benchmark about where the start of this feed may have been. 

speaking of this we should modify follow messages to include the latestSeq for the feed you're acting on. this would be very useful info to have. 

also if you ever mention anyone or dm them that should commend their whole feed for decryption imo

README.md
```diff
@@ -492,6 +492,12 @@ const config = {
      */
     dangerouslyKillFlumeWhenMigrated: false,
 
+    /**
+     * Only try to decrypt box1 messages created before this date
```
staltz, line 496: `before` => `after`!

indexes/private.js
```diff
@@ -23,6 +23,10 @@ module.exports = function (dir, sbot, config) {
   let encrypted = []
   let canDecrypt = []
 
+  let startDecryptBox1 = null
+  if (config.db2.startDecryptBox1)
+    startDecryptBox1 = new Date(config.db2.startDecryptBox1)
```
staltz, line 28: I'd feel a bit safer if `startDecryptBox1` were immutable, so something like

```js
const startDecryptBox1 = config.db2.startDecryptBox1
  ? new Date(config.db2.startDecryptBox1)
  : null
```

staltz:
Just two suggestions and otherwise it looks green!

arj03:
Nice, thanks for the review. Will fix them later today.

Btw. I can have a go add adding this to manyverse in the form of: only for new users and only the first time they start the app (so should be safe), if that sounds good? :)

staltz:
> I can have a go add adding this to manyverse in the form of: only for new users and only the first time they start the app (so should be safe), if that sounds good? :)

Sure! Let me know if you need help with the setup. Easiest way is to build and test the desktop app.

How would the `startDecryptBox1` date be persisted? It needs to be stored in disk right, because it needs to be passed to the db2 config every time on startup. There are other settings that we save, such as `hops` which are loaded into the config too.

And what about recovering the `startDecryptBox1` when the user loses their db and restores their account?

arj03:
Thanks. 

I choose that specific case, because there we don't need to persist anything. And I still think it should be useful. We can always add more cases later if we find them safe.

staltz:
Ah, I see, "only the first time they start the app". Sounds good!

PR #319, opened 2022-03-30
dont prune jit from status object

staltz:
Similar to https://github.com/ssb-ngi-pointer/jitdb/pull/201


arj03 (approved):
Ok, I trust you on this :)

PR #321, opened 2022-03-30
prepare() should not wait for base index

staltz:
## Context

#320 

This is the final piece of the puzzle for stable progress bars in Manyverse, I tested these changes already in the app.

## Problem

`prepare()` is doing `onDrain()` which means we're waiting for the base (leveldb) index to be built *before* starting the JITDB prepare.

Not only is this unnecessary (JITDB prepare is entirely independent from base index), it would also cause a jump backwards in the progress calculations, because first the leveldb indexes would be built (progress bar going forwards), and then once done the status update for JITDB would appear at the start (progress bar going back).

## Solution

Don't do `onDrain()`.

I tried to write a test to catch this use case, but couldn't find a way.

### DIFF
--- a/db.js
+++ b/db.js
@@ -617,10 +617,10 @@ exports.init = function (sbot, config) {
   function prepare(operation, cb) {
     if (sbot.db2migrate) {
       sbot.db2migrate.synchronized((isSynced) => {
-        if (isSynced) onDrain(next)
+        if (isSynced) next()
       })
     } else {
-      onDrain(next)
+      next()
     }
 
     function next() {
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "flumecodec": "0.0.1",
     "flumelog-offset": "3.4.4",
     "hoox": "0.0.1",
-    "jitdb": "^6.1.0",
+    "jitdb": "^6.1.1",
     "level": "^6.0.1",
     "level-codec": "^9.0.2",
     "lodash.debounce": "^4.0.8",


arj03 (approved):
Great. Are you seeing the app is "faster" to show initial content with these prepare? It was always the plan to do something like that, so really glad to see you adding them.

staltz:
> Are you seeing the app is "faster" to show initial content with these prepare?

So far I'm just running the app with ssb-fixtures, looking at the progress bars. I'm gonna test it later in production, on mobile too etc.

staltz:
@arj03 I ran manyverse 0.2203.21 and manyverse on my local branch to compare them. First I delete `db2/indexes`, then I measure time between "splashscreen ended and actual app started" and "I can see the 1st thread on the feed". I ran it 3 times

## Before

- 4m10s
- 4m54s
- 4m28s

## After

- 4m28s
- 4m32s
- 4m44s

So there's not that much difference. I wonder if it makes it slower to build many indexes at the same time, versus doing one index at a time.

I wasn't working on improving performance, so this is fine by me. It's great that the progress bar indeed does not go backwards.

arj03:
The best would probably be to prepare one big query with all the things you need, like author, type etc.

I wonder how long that time would be if you disabled all level indexes inside db2 so it was only jitdb doing its work. Just out of curiosity. It should be around 1 minute.

staltz:
> The best would probably be to prepare one big query with all the things you need, like author, type etc.

I was doing one `prepare()` for each JITDB index, and I switched now to one big `and()` of all of those, but then the duration jumped up to 5m28s. Leveldb indexes were still in place, though.

> I wonder how long that time would be if you disabled all level indexes inside db2 so it was only jitdb doing its work. Just out of curiosity. It should be around 1 minute.

I haven't done this yet

PR #322, opened 2022-03-30
improve progress calculation in status.js

staltz:
See https://github.com/ssb-ngi-pointer/jitdb/pull/204

### DIFF
--- a/db.js
+++ b/db.js
@@ -137,19 +137,8 @@ exports.init = function (sbot, config) {
     state[kvt.value.author] = PrivateIndex.reEncrypt(kvt)
   }
 
-  // Crunch stats numbers to produce one number for the "indexing" progress
   status.obv((stats) => {
-    const logSize = Math.max(1, stats.log) // 1 prevents division by zero
-    const nums = Object.values(stats.indexes).concat(Object.values(stats.jit))
-    const N = Math.max(1, nums.length) // 1 prevents division by zero
-    const progress = Math.min(
-      nums
-        .map((offset) => Math.max(0, offset)) // avoid -1 numbers
-        .map((offset) => offset / logSize) // this index's progress
-        .reduce((acc, x) => acc + x, 0) / N, // avg = (sum of all progress) / N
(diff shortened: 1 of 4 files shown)


staltz:
Putting this in draft because there's another thing I have to address: sometimes the progress bar grows up to ~85% and stays stuck there too. I think it's because there may be replication happening simultaneously to `prepare()` indexes being built, and then when those `prepare()` are done, they are outdated with the latest messages that were replicated. Because it never reached 100%, it didn't `resetStatus()`.

I have a clutch idea: wait 3 minutes and if the status hasn't moved forward, reset it. But I would really like to find a better idea. Thoughts @arj03 ?

arj03:
How about, only include jitdb in progress total if its actually indexing. If not, consider it as 100%, back to the JIT nature. Otherwise we are working against the system I think.

staltz:
Good point, but how can we determine that jitdb is "indexing"? Also, that idea could potentially cause the progress bar to jump backwards if you start with both leveldb and jitdb indexing simultaneously, but then when jitdb is done but leveldb is still ongoing, we would remove jitdb from the calculation, and this would cause the total calculation to jump backwards.

arj03:
Maybe remove was a bad word, include it as 100% so it doesn't jump backwards. I was looking at status.batchUpdate in jitdb and it appears we should be able to add a bit more information if the indexing is active or not when it is called. Like [here](https://github.com/ssb-ngi-pointer/jitdb/blob/caac19b22f20aeef2b6731f7506503fc9817a504/index.js#L578) it is done.

arj03:
Also this kind of information is quite interesting, you can then see when jitdb is running or not so that will make it easier to check what is going on for initial sync and those kind of cases.

staltz:
@arj03 And this PR is ready

arj03 (approved):
Looks good from a quick glance. The most important thing here is that you have been testing this in manyverse and checked that the numbers look good. Who would have thought progress bars could be such a task? :)

PR #323, opened 2022-04-01
calculate progress when log is empty

staltz:
**Problem:** when the log is empty, `stats.jit` is still considered at "100%" while all the leveldb indexes are considered at "0%", so averaging them would give something like "10%" stuck.

**Solution:** if the log is empty, just return 100% progress.


arj03 (approved):
Yep

PR #324, opened 2022-04-03
Support passing in pValue

arj03:
This is what I have used to test https://github.com/ssb-ngi-pointer/jitdb/pull/208. So not really ready yet.

Also we can do the same trick here for level indexes :)

### DIFF
--- a/db.js
+++ b/db.js
@@ -496,7 +496,7 @@ exports.init = function (sbot, config) {
       paused: false,
       write(record) {
         const buf = record.value
-        const pValue = bipf.seekKey(buf, 0, B_VALUE)
+        const pValue = buf ? bipf.seekKey(buf, 0, B_VALUE) : null
         indexesArr.forEach((idx) => idx.onRecord(record, false, pValue))
       },
       end() {
@@ -510,7 +510,7 @@ exports.init = function (sbot, config) {
             paused: false,
             write(record) {
               const buf = record.value
-              const pValue = bipf.seekKey(buf, 0, B_VALUE)
+              const pValue = buf ? bipf.seekKey(buf, 0, B_VALUE) : null
               indexesArr.forEach((idx) => idx.onRecord(record, true, pValue))
             },
           })
--- a/db.js
+++ b/db.js
@@ -490,10 +490,14 @@ exports.init = function (sbot, config) {
     )
     debug(`lowest offset for all indexes is ${lowestOffset}`)
 
+    const B_VALUE = Buffer.from('value')
+
     log.stream({ gt: lowestOffset }).pipe({
       paused: false,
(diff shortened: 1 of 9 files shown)


arj03:
Some numbers:

```
Only base + keys index & query for 1 author empty indexes

base: 41.1s
no decrypt: 30.2
no decrypt + jitdb pValue: 25.6

Base, keys, ebt, full mentions index & query for 1 author empty indexes

base: 56.3s
no decrypt + jitdb pValue: 37.7
```

So the more level plugins you have, the better it gets. It might also be a good idea to pass things like pContent to level plugins. Maybe even jitdb.

db.js
```diff
@@ -490,10 +490,14 @@ exports.init = function (sbot, config) {
...
     debug(`lowest offset for all indexes is ${lowestOffset}`)
 
+    const B_VALUE = Buffer.from('value')
+
     log.stream({ gt: lowestOffset }).pipe({
       paused: false,
       write(record) {
-        indexesArr.forEach((idx) => idx.onRecord(record, false))
+        const buf = record.value
+        const pValue = bipf.seekKey(buf, 0, B_VALUE)
```
staltz, line 499: We could move these two lines into the internals of `idx.onRecord` (indexes/plugin.js). It's the right place where to do this, and it avoids the duplicated code (lines 512-513 below)

indexes/about-self.js
```diff
@@ -40,14 +39,13 @@ module.exports = class AboutSelf extends Plugin {
     )
   }
 
-  processRecord(record, seq) {
+  processRecord(record, seq, pValue) {
+    if (pValue < 0) return
```
staltz, line 43: Could we move this `if` into `onRecord` (indexes/plugin.js)? I'm thinking that since this is ssb-db2, it can assume it knows about SSB msg formats, and I think in all records we have KVTs, right?

staltz (changes requested):

db.js
arj03: But then each plugin would have to calculate them?

staltz: No, because `onRecord` is a generic wrapper over each plugin's `processRecord` method

arj03: I still don't understand :-) There is a foreach loop on the next line. Unless we refactor the whole write into a function.

staltz (approved):
Oh silly me, now I see why we can't move the pValue calculation. Sorry for that :)

staltz:
@arj03 By the way, I just had an idea for using WeakMap so that we don't need to recalculate *any* of these pointers. It could be put into BIPF and would work transparently. Then we could revert these pValue changes and we would still have the same performance benefits

arj03:
Uhh nice!

PR #328, opened 2022-04-06
improve seekers so they always return numbers

staltz:
Similar to https://github.com/ssb-ngi-pointer/jitdb/pull/211

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "flumecodec": "0.0.1",
     "flumelog-offset": "3.4.4",
     "hoox": "0.0.1",
-    "jitdb": "^6.4.0",
+    "jitdb": "^6.4.1",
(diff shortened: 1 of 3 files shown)


arj03:
Migration test seems to be failing? test/migration.js:353

staltz:
Yeah I'm looking into it

staltz:
I don't know what was wrong before, but `50` is the correct number, I ran it manually on paper. Maybe this PR made the code behave more correctly and the tests were wrong?

See the comment I added in the tests, inspired by https://github.com/ssb-ngi-pointer/ssb-fixtures/blob/249d6965b8a69f2ed160dba8872a46bced89b386/test/index-feeds.js#L28-L29

arj03 (approved):
Yeah, great to have these cleaned up

PR #329, opened 2022-04-08
add code coverage

staltz:
@arj03 This coverage allowed me to check that we don't have currently tests for `reindexOffset`, and that it seems like we need to update the implementation anyway, because it calls `processRecord` (not `onRecord`) and doesn't pass pValue. I'll work on that next.


arj03 (approved):
Ah, good find :)

PR #330, opened 2022-04-08
force flush after reindexEncrypted is done

staltz:
## Context

Working on https://github.com/ssb-ngi-pointer/jitdb/issues/199

## Problem

There are no tests for reindexEncrypted in this repo, and as a result we were using the `processRecord` API incorrectly (not passing pValue arg), and even though we processed records on leveldb indexes, the changes were not persisted to disk.

## Solution

I copied the test from ssb-db2-box2, but tweaked it to make sure that we query a leveldb index before and after reindexEncrypted. Also, after reindexOffset is called for all leveldb indexes, force flush to disk. 

### DIFF
--- a/db.js
+++ b/db.js
@@ -39,6 +39,11 @@ const makeBaseIndex = require('./indexes/base')
 const KeysIndex = require('./indexes/keys')
 const PrivateIndex = require('./indexes/private')
 
+const B_VALUE = Buffer.from('value')
+const B_KEY = Buffer.from('key')
+const B_META = Buffer.from('meta')
+const B_PRIVATE = Buffer.from('private')
+
 const { where, fromDB, key, author, deferred, toCallback, asOffsets } =
   operators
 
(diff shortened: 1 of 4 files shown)


db.js
```diff
@@ -490,27 +495,25 @@ exports.init = function (sbot, config) {
...
     debug(`lowest offset for all indexes is ${lowestOffset}`)
 
-    const B_VALUE = Buffer.from('value')
-
     log.stream({ gt: lowestOffset }).pipe({
       paused: false,
       write(record) {
         const buf = record.value
-        const pValue = buf ? bipf.seekKey(buf, 0, B_VALUE) : null
+        const pValue = buf ? bipf.seekKey(buf, 0, B_VALUE) : -1
```
arj03, line 502: 👍

arj03 (approved):
Looks good. Good with extra tests here.

PR #332, opened 2022-04-08
fix usage of prettier-ignore

staltz:
Housekeeping PR.

Turns out we can't use `// prettier-ignore` at the end of the line, it makes prettier error/crash and then it doesn't do any actual code formatting.

It has to be on its own line, before the statement.


arj03 (approved):
ugh. Tyranny of the pretty printer

PR #336, opened 2022-04-18
improve performance of del() by 20x

staltz:
Uses `jitdb.lookup()` (see https://github.com/ssb-ngi-pointer/jitdb/issues/215) to improve the performance of `del()` and `getMsg()`.

See the numbers for `del()`:

## Before

|  | Duration |
|--|--|
| Publish 1000 messages | 359.185ms |
| Delete 500 messages | 1039.531ms |

## After

|  | Duration |
|--|--|
| Publish 1000 messages | 350.658ms |
| Delete 500 messages | 36.623ms |
| Wait for deletes to flush | 264.364ms |

Note that the flushing of deletes (264ms) is probably dominated by the duration of the [debounce period which is 250ms](https://github.com/ssb-ngi-pointer/async-append-only-log/blob/bec87f28c0f9a3e165fd36f98c8d3ba5622a40d4/index.js#L277).

### DIFF
--- a/db.js
+++ b/db.js
@@ -180,10 +180,10 @@ exports.init = function (sbot, config) {
   function getHelper(id, onlyValue, cb) {
     if (sbot.db2migrate) {
       sbot.db2migrate.synchronized((isSynced) => {
-        if (isSynced) onDrain(next)
+        if (isSynced) onDrain('keys', next)
       })
     } else {
-      onDrain(next)
+      onDrain('keys', next)
     }
 
     function next() {
--- a/db.js
+++ b/db.js
@@ -478,7 +478,15 @@ exports.init = function (sbot, config) {
     const guard = guardAgainstDuplicateLogs('del()')
     if (guard) return cb(guard)
 
-    onDrain('keys', () => {
+    if (sbot.db2migrate) {
+      sbot.db2migrate.synchronized((isSynced) => {
+        if (isSynced) onDrain('keys', next)
+      })
+    } else {
+      onDrain('keys', next)
+    }
+
+    function next() {
       indexes['keys'].getSeq(msgId, (err, seqStr) => {
         if (err)
           return cb(clarify(err, 'del() failed to find msgId from index'))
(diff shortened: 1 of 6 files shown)


db.js
```diff
@@ -539,10 +576,9 @@ exports.init = function (sbot, config) {
       indexName = 'base'
     }
 
-    // setTimeout to make sure extra indexes from secret-stack are also included
-    setTimeout(() => {
+    secretStackLoaded.onReady(() => {
```
staltz, line 579: Notice that the `setTimeout` was causing a forced async wait for **every** onDrain call we did, even if the setTimeout was useful only during init (to wait for secret-stack plugins to register indexes). This new method (using `ReadyGate`) means that if it's `ready`, it'll be called immediately/synchronously. I expect this will improve overall all ssb-db2 queries a tiny bit.

utils.js
```diff
@@ -26,4 +26,22 @@ function onceWhen(obv, filter, cb) {
   })
 }
 
-module.exports = { onceWhen }
+class ReadyGate {
```
staltz, line 29: I made this class because I wanted "synchronous deferred promises", but no such thing exists. The internals of this class may be familiar to you, we use this kind of technique in jitdb: https://github.com/ssb-ngi-pointer/jitdb/blob/8f3b79abaf2c90b673323f461dd9abd3910a0a88/index.js#L41-L42

db.js
arj03: Yes, great find

arj03 (approved):
Can't see anything wrong with this. It looks solid. I checked `get` and can see it's only really used in `addOOO` inside this module. I went checking ssb-threads and it does call getMsg potentially quite a few times. I could be interesting to add a console.log to this getMsg and fire up Manyverse to see how many times that method is called within say the first 1 minute of "normal" usage.

arj03:
Great, so it should also help a tiny bit overall when using ssb-threads. Great to hear! :)

PR #337, opened 2022-04-19
update our usage of Obz to be safer

staltz:
This PR does 3 refactors.

I started by updating Obz to 1.1.0 to use the new `return false` style, but I noticed that actually our util `onceWhen` is safer (because if we were to use `return false` we have to run that as the last command, but in fact we need to unsubscribe from the Obz as soon as possible). So there are more things that use `onceWhen` now.

The other refactors are simple:

- replace `forEach()` with `for (const x of ..` because it's a bit better when debugging stack traces
- replace Promise.all with `multicb` where possible (because it's "quicker", promises add a microqueue task delay)

### DIFF
--- a/db.js
+++ b/db.js
@@ -179,9 +179,11 @@ exports.init = function (sbot, config) {
 
   function getHelper(id, onlyValue, cb) {
     if (sbot.db2migrate) {
-      sbot.db2migrate.synchronized((isSynced) => {
-        if (isSynced) onDrain('keys', next)
-      })
+      onceWhen(
+        sbot.db2migrate.synchronized,
+        (isSynced) => isSynced === true,
+        () => onDrain('keys', next)
+      )
     } else {
       onDrain('keys', next)
     }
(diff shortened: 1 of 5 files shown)


db.js
```diff
@@ -554,21 +558,22 @@ exports.init = function (sbot, config) {
...
       },
       end() {
         debug(`updateIndexes() scan time: ${Date.now() - start}ms`)
-        const flushTasks = indexesArr.map((idx) =>
-          promisify(idx.flush.bind(idx))()
-        )
-        Promise.all(flushTasks).then(() => {
+        const doneFlushing = multicb({ pluck: 1 })
+        for (const idx of indexesArr) idx.flush(doneFlushing())
+        doneFlushing((err) => {
```
staltz, line 567: As a bonus, now we're handling errors from flushing indexes, while before we weren't.

db.js
```diff
@@ -671,7 +681,7 @@ exports.init = function (sbot, config) {
         if (idx.indexesContent()) idx.processRecord(record, seq, pValue)
       }
 
-      cb(null, record.offset)
+      cb()
```
staltz, line 684: The result seemed like it was never used?

arj03: Right

arj03 (approved):
Easy review. Looks good

Issue #306, opened 2022-01-19
How to defrag?


staltz:
I've been thinking about getting back to implementing partial replication (finishing https://github.com/ssb-ngi-pointer/ssb-replication-scheduler/pull/5), and putting it in Manyverse.

**TL;DR how can we keep the sizes of log + indexes somewhat _constant/stable_ as M new messages come in and M older messages get overwritten by zero bytes?** 

Deleting messages from the log and defragmenting will be an important part, because initially the use of partial replication would add N index messages to the log, and N could be significant (it's at least the number of votes/posts/contacts/abouts), which could bring the log size from the current ~1GB to 2GB or more. Defragging is important because if we just overwrite the "deleted" records with zero bytes, this still doesn't change the total size of the log, and doesn't change the offsets in the indexes, and there's a limit of ~4.3GB for the max offset that indexes can point to (due to uint32).

I thought about using sparse files for the log, or compressing the log on writes and decompressing on reads, but those would still not fix the problem which is about the uint32 limit for offsets in the _indexes_, not about the log.

So we would need to defrag the log itself, which means also that all indexes have to be rebuilt to match the new offsets for each record. How often to do this defrag+reindex would be tricky, and the waiting experience would be pretty bad for users.

One idea is to upgrade the offsets from uint32 to uint64 which would increase the "lifespan" of the log by 4 million times. Assuming it takes 1 year to get 1GB worth of messages on the average user's log, then we would have millions of years, so uint64 is well enough.

Perhaps if offset variables are updated to uint64, and if we use compression so that the log file itself can shrink the space taken by those zero bytes, then this problem might be "solved". But I'm not sure about that, because index sizes could still keep growing.

Ideally (just to make sure our system works at scale), we should assume that an insane amount of new messages come in via replication (say, +1GB on the log every month. This is much more that the previous mentioned +1GB/year, but the "lifespan" of the log with uint64 would still be ridiculous, about 350k years), and that an equal amount of older messages gets "deleted" (overwritten by zero bytes), and how to keep the user's disk usage somewhat stable. I mean, it's acceptable if the disk usage has a sawtooth pattern (like garbage collection sawtooth charts), as long as an upper bound is never crossed, i.e. no "disk" leaks happen.

arj03:
Some loose thoughts (I'll need to spend more time on it). 

[...]

PR #339, opened 2022-04-20
add compact() API and auto-rebuild indexes

staltz:
Context: https://github.com/ssb-ngi-pointer/ssb-db2/issues/306 and https://github.com/ssb-ngi-pointer/jitdb/issues/199

This PR adds the `compact()` ("async") API to ssb-db2 which just triggers async-append-only-log `compact()`. When it's done, it will rebuild all leveldb indexes from scratch, rebuild private indexes, and partially rebuild jitdb indexes. This is all done in a way that is crash-resistant, capable of resuming where it had stopped. And compaction is guaranteed to not happen concurrently with log.stream that builds indexes, or any jitdb query.

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,16 +224,17 @@ sbot.db.getIndex('myindex').myOwnMethodToGetStuff()
 Or you can wrap that in a secret-stack plugin (in the example above,
 `exports.init` should return an object with the API functions).
 
-There are other special methods you can implement in order to add
-"hooks" in the `Plugin` subclass:
+There are other optional methods you can implement in the `Plugin` subclass:
 
-- `onLoaded(cb)`: called once, at startup, when the index is
-  successfully loaded from disk and is ready to receive queries
-- `onFlush(cb)`: called when the leveldb index is about to be saved to
+- `onLoaded(cb)`: a hook called once, at startup, when the index is successfully
+  loaded from disk and is ready to receive queries
+- `onFlush(cb)`: a hook called when the leveldb index is about to be saved to
   disk
-- `indexesContent()`: method used when reindexing private group
-  messages to determine if the leveldb index needs to be updated for
-  decrypted messages. The default method returns true.
+- `indexesContent()`: method used when reindexing private group messages to
+  determine if the leveldb index needs to be updated for decrypted messages. The
+  default method returns true.
+- `reset()`: a method that you can use to reset in-memory state that you might
+  have in your plugin, when the leveldb index is about to be rebuilt.
 
 ### Compatibility plugins
 
@@ -268,7 +269,7 @@ const sbot = SecretStack({ caps })
 The following is a list of modules that works well with ssb-db2:
 
  - [ssb-threads] for working with post messages as threads
- - [ssb-suggest-lite] for fetching profiles of authors 
+ - [ssb-suggest-lite] for fetching profiles of authors
  - [ssb-friends] for working with the social graph
  - [ssb-search2] for full-text searching
  - [ssb-crut] for working with records that can be modified
@@ -463,7 +464,7 @@ Use [JITDB's prepare](https://github.com/ssb-ngi-pointer/jitdb/#prepareoperation
 
 Waits for the index with name `indexName` to be in sync with the main
 log and then call `cb` with no arguments. If `indexName` is not
-provided, the base index will be used. 
+provided, the base index will be used.
 
 The reason we do it this way is that indexes are updated
 asynchronously in order to not block message writing.
--- a/README.md
+++ b/README.md
@@ -469,6 +469,11 @@ provided, the base index will be used.
 The reason we do it this way is that indexes are updated
 asynchronously in order to not block message writing.
 
+### compact(cb)
+
+Compacts the log (filling in the blanks left by deleted messages and optimizing
+space) and then rebuilds indexes.
+
 ## Configuration
 
 You can use ssb-config parameters to configure some aspects of ssb-db2:
--- a/db.js
+++ b/db.js
@@ -99,6 +99,7 @@ exports.init = function (sbot, config) {
   const post = Obv()
   const indexingProgress = Notify()
   const indexingActive = Obv().set(0)
+  let abortLogStreamForIndexes = null
   const compacting = Obv().set(false)
   const hmacKey = null
   const stateFeedsReady = Obv().set(false)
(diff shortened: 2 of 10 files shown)


staltz:
There's an error in the test `compaction resumes automatically after a crash`:

```
RangeError [ERR_OUT_OF_RANGE]: The value of "offset" is out of range. It must be >= 0 and <= 65534. Received 77082
    at new NodeError (internal/errors.js:322:7)
    at boundsError (internal/buffer.js:85:9)
    at Buffer.readUInt16LE (internal/buffer.js:242:5)
    at Object.readDataLength (/home/runner/work/ssb-db2/ssb-db2/node_modules/async-append-only-log/record.js:23:19)
    at Object.getDataNextOffset (/home/runner/work/ssb-db2/ssb-db2/node_modules/async-append-only-log/index.js:234:31)
    at Stream._handleBlock (/home/runner/work/ssb-db2/ssb-db2/node_modules/async-append-only-log/stream.js:89:37)
    at Stream._resumeCallback (/home/runner/work/ssb-db2/ssb-db2/node_modules/async-append-only-log/stream.js:145:27)
    at Object.getBlock (/home/runner/work/ssb-db2/ssb-db2/node_modules/async-append-only-log/index.js:201:7)
    at Stream._resume (/home/runner/work/ssb-db2/ssb-db2/node_modules/async-append-only-log/stream.js:136:12)
    at Stream._next (/home/runner/work/ssb-db2/ssb-db2/node_modules/looper/index.js:11:9) {
  code: 'ERR_OUT_OF_RANGE'
}
```

It happens mostly in CI, hard to reproduce locally. Have to solve this crash before the PR can be said to be ready.

staltz:
I fixed [this error](https://github.com/ssb-ngi-pointer/ssb-db2/pull/339#issuecomment-1103988874), it was about `updateIndexes()` (and a `log.stream()`) happening concurrently with compaction.

When running `compact()` in ssb-db2, it checks whether there is any active `updateIndexes()` going on, postponing the compact if there are.

However, if we are resuming from a crash, then compaction is started *automatically* by async-append-only-log, and then `updateIndexes()` is also started automatically. So I fixed this by adding an if condition in `updateIndexes()` which checks whether compaction is ongoing.

db.js
```diff
@@ -731,13 +775,73 @@ exports.init = function (sbot, config) {
...
+            if (err) console.error('ssb-db2 reindex jitdb after compact', err)
+            rimraf.sync(reindexJitPath(dir))
+          })
+        }
+        compactStartOffset = null
+      } else {
+        rimraf.sync(resetLevelPath(dir))
+        rimraf.sync(resetPrivatePath(dir))
+        rimraf.sync(reindexJitPath(dir))
+        restartUpdateIndexes()
```
staltz, line 840: This `restartUpdateIndexes()` addresses the following case:

- `log.compact()` was started immediately, before `updateIndexes()` could start
- Crash
- Resume
- compaction is resumed automatically by AAOL
- `updateIndexes()` is cancelled because compaction is ongoing
- compaction is done, but `sizeDiff === 0`
- **At this point, leveldb indexes are outdated** (they haven't had the chance to build at all!) but there is no `updateIndexes()` ongoing, so we have to kickstart it somehow

staltz:
@arj03 Although this is the final task for ssb-db2, I have still 2 TODOs:

- EBT: we need to update or delete some entries after we delete msgs from the log. To keep it simple, let's suppose we only use sbot.db.deleteFeed and thus we delete that whole feedId from EBT too
- What about browser compatibility for AAOL/compaction.js and DB2 post-compact files?

db.js
```diff
@@ -731,13 +775,73 @@ exports.init = function (sbot, config) {
...
+  }
+
+  function compact(cb) {
+    if (notYetZero(jitdb.indexingActive, compact, cb)) return
+    if (notYetZero(jitdb.queriesActive, compact, cb)) return
+    if (notYetZero(indexingActive, compact, cb)) return
+
+    fs.closeSync(fs.openSync(resetLevelPath(dir), 'w'))
+    fs.closeSync(fs.openSync(resetPrivatePath(dir), 'w'))
+    fs.closeSync(fs.openSync(reindexJitPath(dir), 'w'))
```
arj03, line 802: Yes, this is related to:

> What about browser compatibility for AAOL/compaction.js and DB2 post-compact files

I'll have to test this to be sure. Currently writing a spec for the new feed format, so will circle back to this later. For now I think this shouldn't be a show-stopper for this PR.

staltz: Maybe could change these to atomic-file-rw

db.js
```diff
@@ -774,7 +879,6 @@ exports.init = function (sbot, config) {
     getState: () => state,
     getIndexes: () => indexes,
     getIndex: (index) => indexes[index],
-    clearIndexes,
```
arj03, line 777: This was something I used in browser core for testing. I don't mind removing this. It's a thin wrapper around indexes reset anyway.

db.js
arj03: Well there is also something in AAOL anyway so have to check both. So if you tested this and it works, fine to leave it.

db.js
```diff
@@ -21,7 +22,6 @@ const pull = require('pull-stream')
 const paramap = require('pull-paramap')
 const Ref = require('ssb-ref')
 const Obv = require('obz')
-const promisify = require('promisify-4loc')
```
arj03, line 24: this should also be removed from package.json

db.js
arj03: I was looking at this to make sure that it works correctly in a crash and I think it does because:

You don't need to call reset on indexes because if that was needed (sizeDiff > 0) then we would already know because this info is persisted in AAOL and would not disappear in a crash.

In other words: 👍

README.md
```diff
@@ -463,7 +464,7 @@ Use [JITDB's prepare](https://github.com/ssb-ngi-pointer/jitdb/#prepareoperation
 
 Waits for the index with name `indexName` to be in sync with the main
 log and then call `cb` with no arguments. If `indexName` is not
-provided, the base index will be used. 
+provided, the base index will be used.
```
arj03, line 467: can you add `compact` to this document under methods?

staltz: Oh, yes!

db.js
staltz: Do you want to bring it back?

arj03 (approved):
Nice work 🚀

staltz:
Feels crazy to finally merge this, but here we go!

PR #343, opened 2022-05-16
Butt2

arj03:
Almost ready, need to do something about `buttPost`.


staltz:
@arj03 Thanks! I'm actually not sure which way will be easier: merging this now, or resolving conflicts later. It might be complicated any way. If you're not blocked, then maybe let me try to do the format&encryption PR first, and then if resolving conflicts here is too hard, it's not going to be hard for me to write a new PR to add buttwoo using the new format&encryption API.

> About the node 12 and 14 tests not working, I think it might be related to npm install never finishing

Oh yeah, that one is still a mystery. It actually does complete, if you would wait a long time (CI was set to maximum 10min). And it's not related to `uuid`, try running `npm install --verbose` and you'll see where it gets slow. It's related to the repos we moved from ssb-ngi-pointer to ssbc but I don't know *why*.

db.js
```diff
@@ -863,6 +913,9 @@ exports.init = function (sbot, config) {
     reindexEncrypted,
     indexingProgress: () => indexingProgress.listen(),
 
+    buttPost,
```
staltz, line 916: Can you help me understand what is `buttPost` for? Why isn't `post` enough? (I'm thinking and planning whether we need `post` for each feed format)

arj03: Post worked for bendy butt because we made it behave very similar to a normal msgval. If we really want to do binary formats, then we need something where we can give the raw bytes. So i added this. I was thinking we should have one post per feed type. Then in EBT you don't need to test so much. You already know the format.

arj03: Not really one variable per type. Maybe a map

db.js
```diff
@@ -426,6 +429,53 @@ exports.init = function (sbot, config) {
...
+          return cb('incorrect buttwoo format, addButtwoo() failed')
+        if (data[0].length !== 3 || data[1].length !== 8)
+          return cb('incorrect buttwoo format, addButtwoo() failed')
+
+        const msgKeyBFE = butt2.hash(data)
+        const msgBIPF = butt2.butt2ToBipf(data, msgKeyBFE)
+
+        const author = bfe.decode(data[1][0])
+        const parent = bfe.decode(data[1][1])
+        const authorParent = author + (parent === null ? '' : parent)
```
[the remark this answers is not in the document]
arj03: this is for subfeeds. the top feed would have parent nil. So that works like any other feed, but for subfeeds this is used to distinguish them as they would use the same sequences as another feed for this feed id. it might be good to probably only do this for buttwoo messages

db.js
staltz: Okay, thanks for the context. I have a few options how to take that into account. I'm thinking of renaming (or maybe not renaming) `post` to `onAdded` or `onMsgAdded` and then besides passing the msg (bytes or object), it would also tell what format it was. That would inform EBT.

db.js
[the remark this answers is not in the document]
staltz: This is confusing, can you help me understand? I didn't find anything in the tests in `ssb-buttwoo` for this, and neither in this `ssb-db2` PR.

Suppose a "metafeed" in buttwoo has feed Id `ssb:feed/buttwoo-v1/TBeL1=`. Does this mean that a subfeed of that one has `ssb:feed/buttwoo-v1/TBeL1=` also as the same feedId? Also, what is even "parent"? Is it the feedId of the parent metafeed, or is a message Id? (For what message??)

And why is `authorParent` here only used internally in the codebase? Why couldn't the feedId for this subfeed be the SSB URI `ssb:feed/buttwoo/<FEEDID>/<PARENT>`?

I'll keep asking until I understand it properly, because I think it's important more than one person understands this.

For the `addFeedFormat()` refactor I really need to understand this, because buttwoo works differently than classic and bendy butt and the mechanism I come up with has to work well for buttwoo too.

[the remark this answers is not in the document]
arj03: First of all, it is great you are asking these kind of questions. This module and ebt is where we really see how the format plays out :)

I updated the spec, especially around subfeeds to make this more clear (I hope) also what parent is. The big distinction is that feedId and author are potentially different and in classic they are always the same. If we used `ssb:feed/buttwoo/<FEEDID>/<PARENT>` then for the implementation it would be a lot easier, because we would just have 1 field, but then one of the appeals of this, compared to bendy butt would go away: that for example about message would have the same author as the normal messages, so they are easy to look up. That said, some kind of separator between the author and parent would probably be good.

As for tests, the great good thing about the current scheme is that buttwoo doesn't have to care very much about this, you just need to make sure that you supply the correct previous (compared to the feed, not author) and this is really the job of db2 in our stack. So a test case here would be good, I did test it in ebt where we also need to handle this author vs feed in a special way.

All this said, it's clear that subfeeds does complicate things. Besides the changes in ebt & db2, the replication scheduler also needs to know about these. I do really like the idea, but it comes back to how we roll things out. For private groups we'll use and properly test metafeeds. These could be using buttwoo, but would not use the subfeeds feature. Then there is if/when we change the main feed. Here I could see it might make sense, but then again, we already have all the functionality in place to look up things like about messages given a bendy butt subfeed. What I'm saying is, I'm not against removing subfeeds from the spec to keep things simpler. If we do that I'll probably keep the design around under a different name. In any case, I'm curious to hear what you think :)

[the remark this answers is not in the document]
staltz: Okay, we're making progress (in my understanding, haha)! Thanks for this. I still don't understand some things, but I can guess what they are doing. I'm going to make a PR to the buttwoo spec and you can let me know if my understanding there is correct or not.

db.js
```diff
@@ -426,6 +429,53 @@ exports.init = function (sbot, config) {
...
+          previous.key,
+          hmacKey
+        )
+
+        if (err)
+          return cb(clarify(err, 'message validation in addButtwoo() failed'))
+
+        state[authorParent] = {
+          key: msgKeyBFE,
+          value: data,
```
staltz, line 466: @arj03 One thing I bumped into, and maybe this is new info for you: here we are setting the in-memory `state` to `authorParent => [encodedValue, sig, encodedContent]`, because `data` is that array with 3 parts.

However, when loading ssb-db2 from a cold start, the "latest" from Base index is loaded with KVTs (JS objects) from the log. So it's `author => KVT` that we'll have in `state`. Different shape for the state dictionary.

This means that after a cold start, if you call `addButtwoo`, the line 456 has a JS KVT (In fact, it's always going to have `null` as the previous.value, because the `state` is loaded with `author` from cold start, not `authorParent`). But subsequent `addButtwoo` calls will have an array with 3 parts on line 456.

In other words, the shape of `state` is not consistent here.

arj03: Ah, we need a test for that. Good spotting.

PR #345, opened 2022-05-26
Faster ebt for encrypted messages

arj03:
There is no need to first decrypt the message and then reEncrypt them ;)

There is a test [here](https://github.com/ssbc/ssb-db2/blob/c35306e265ee67b60a87a355a992ebd9880d6f90/test/ebt.js#L78) for that case. Just so we are sure this is not breaking anything.

### DIFF
--- a/indexes/ebt.js
+++ b/indexes/ebt.js
@@ -5,7 +5,6 @@
 const bipf = require('bipf')
 const clarify = require('clarify-error')
 const Plugin = require('./plugin')
-const { reEncrypt } = require('./private')
 
 const BIPF_AUTHOR = bipf.allocAndEncode('author')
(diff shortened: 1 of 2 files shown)


staltz (approved):
Smart!

CI is an annoying state, I wish I knew how to fix it. I installed and tested with node 12.x locally just to be sure. It's passing :+1:

arj03:
Benchmarks here for comparison. This is for a feed where all messages are encrypted:

```
# classic performance before
ok 1 wait for replication to complete
after 3000 ms bob has 6336
^C
arj@arj:~/dev/ssb-ebt$ node test/formats.js
TAP version 13
# classic performance after
ok 1 wait for replication to complete
after 3000 ms bob has 8598
```

PR #347, opened 2022-06-07
feedFormats and encryptionFormats as plugins

staltz:
## Context

We are adding [buttwoo](https://github.com/ssbc/ssb-buttwoo) as a feed format, and it works differently than other formats, so we would introduce a lot of complexity. At the same time, we need to improve the situation of support for box2 before we move on with "ssb-tribes2" on ssb-db2.

## Solution

**`sbot.db.installFeedFormat()` and `sbot.db.installEncryptionFormat()` and _encodings_.**

These are now orthogonal concerns, so *any* encryption format can work with *any* feed format, which can support different encodings.

### :large_blue_circle:  Encoding

An "encoding" is a way of representing a feed message. This PR supports two encodings: `'js'` (JavaScript objects) and `'bipf'` (BIPF buffers).

### :yellow_circle: Feed Format

A feed format defines how to create messages that follow a schema and a spec. Every feed format is a plugin-like object with:

- `name` string
- `encodings` array of supported encoding names
- functions to create and convert "native messages"
  - `newNativeMsg(opts)`
  - `toNativeMsg(msgVal, encoding)`
  - `fromNativeMsg(nativeMsg, encoding)`
  - `fromDecryptedNativeMsg(plaintextBuf, nativeMsg, encoding)`
- encryption-related function
  - `toPlaintextBuffer(opts)`
- helper functions `getX()` and `isX()`
  - `getFeedId(nativeMsg)` 
  - `getMsgId(nativeMsg)`
  - `isNativeMsg(x)`
  - `isAuthor(author)`
- validation functions
  - `validateSingle()`
  - `validateBatch()`
  - `validateOOOBatch()`

:star: **"Native Msg"** this concept is new and important. Every feed format has a "native message format", which is a shape for the message that is specific to the feed format. The feed format fully controls its own "nativeMsg" shape, and it can be anything.

- For "classic", a nativeMsg is the traditional JavaScript **object** `msgVal`.
- For "bendybutt-v1", a nativeMsg is a BFE-and-bencoded **buffer** with a shape determined by [bendy-butt-spec](https://github.com/ssb-ngi-pointer/bendy-butt-spec).
- For "buttwoo-v1", a nativeMsg is a BFE-and-bipf **buffer** with a shape determined by [buttwoo-spec](https://github.com/ssbc/ssb-buttwoo-spec/). 

### :black_circle:  Encryption Format

An encryption format specifies how to encrypt and decrypt JavaScript buffers intended for some specific recipients. Every encryption format is a plugin-like object with:

- `name` string, also used for the suffix on the ciphertext
- `setup(config, cb)` OPTIONAL 
- `encrypt(plaintextBuf, opts) => ciphertextBuf`
- `decrypt(ciphertextBuf, opts) => plaintextBuf`

## TODO

- [x] private.js special treatments of box1 and box2
- [x] Implement publish in terms of create()
- [x] `isPrivate(encryptionFormat)` for querying canDecrypt
- [x] Move `post` Obv to compat
- [x] :wrench: `feedFormat.getSequence()` (needed for ebt)
- [x] :detective: Is buttwoo performant with ssb-ebt?
- [x] :detective: WeakMap optimizations in feed formats, from/to native msg
- [x] :detective: Figure out box2 key management/addition
- [x] :detective: jitdb operator for querying encrypted-box1 and encrypted-box2
- [x] :wrench: Restore reindex-encrypted.js tests
- [x] :wrench: ssb-db2 should wait for all encryption formats to load
- [x] :wrench: Fix benchmarks
- [x] :abc: Better name for `encryptionFormat.getRecipients()` => `getEncryptionKeys()`
- [x] :abc: `canDecrypt.index` naming?
- [x] :balance_scale: ssb-keyring LGPL-3.0 release
- [x] :rocket: Publish ssb-feed-format
- [x] :rocket: Publish ssb-encryption-format
- [x] :rocket: Publish ssb-classic
- [x] :rocket: Breaking change ssb-bendy-butt
- [x] :rocket: Publish ssb-box1
- [x] :wrench: ssb-uri-spec buttwoo URI needs parent
- [x] :wrench: ssb-uri2 buttwoo URI needs parent
- [x] :rocket: Breaking change ssb-buttwoo
- [x] :book: Document new APIs such as:
  - `create()`
  - `onMsgAdded()`
  - `installFeedFormat()`
  - `installEncryptionFormat()`
- [x] :book: Document breaking changes such as:
  - `encrypted.index`/`canDecrypt.index`
  - `setPost` hack for partial replication in the browser
  - `isPrivate()` => `isDecrypted()` / `isEncrypted()`
- [x] :wrench: ssb-keyring changes to support derived-keys-only workflow
- [x] :rocket: Publish ssb-box2
- [ ] MERGE AND PUBLISH VERSION 5.0.0
- [ ] **ssb-ebt** Replace `post` with `onMsgAdded`
- [ ] **ssb-conn** Replace `post` with `onMsgAdded`
- [ ] **ssb-meta-feeds** Update ssb-bendy-butt to new feed format API
- [ ] ssb-groups triggers reindexEncrypted when a new key is added
- [ ] ssb-groups uses ssb-db2 to query the log for messages and uses ssb-box2 to add keys
- [ ] ssb-groups has helper functions to interact with groups, like ssb-tribes does


staltz:
@arj03 All tests pass except the ones related to box2 simply because of key management that I haven't worked on yet. But box2 is supported and you can see that from `test/create.js`.

To use this in ssb-ebt for buttwoo, you'll have to do something like:

## db => network

- `sbot.getAtSequenceNativeMsg(arr, (err, nativeMsg) =>` (no need to convert, just directly pass the `nativeMsg` to the network peer)

## network => db

- `sbot.db.add(nativeMsg, {encoding: 'bipf', feedFormat: 'buttwoo'}, cb)`
- OR
- `sbot.db.add(nativeMsg, {encoding: 'bipf'}, cb)` (might be just slightly slower than the above)

If you need some other data, let me know.

encryption-formats/box2.js
```diff
@@ -0,0 +1,189 @@
+// SPDX-FileCopyrightText: 2022 Andre 'Staltz' Medeiros
+//
+// SPDX-License-Identifier: LGPL-3.0-only
+
+const BFE = require('ssb-bfe')
+const Ref = require('ssb-ref')
+const { isFeedSSBURI } = require('ssb-uri2')
+const { keySchemes } = require('private-group-spec')
+const { box, unboxKey, unboxBody } = require('envelope-js')
+const { directMessageKey, SecretKey } = require('ssb-private-group-keys')
```
arj03, line 10: these 3 needs to be added to package.json

staltz: Done!

arj03:
I should not add more things here, it's getting late. Good news is that classic replication is working and is around same speed as before :)

feed-formats/buttwoo.js
```diff
@@ -0,0 +1,361 @@
...
+        if (Buffer.compare(parentBFE, parentBFEPrev) !== 0)
+          return cb(new Error('Parent does not match previous message'))
+
+        if (sequence !== sequencePrev + 1)
+          return cb(new Error('Sequence must increase'))
+
+        if (timestamp <= timestampPrev)
+          return cb(new Error('Timestamp must increase'))
+
+        if (Buffer.compare(previousBFE, previousKeyBFE) !== 0)
```
arj03, line 342: `previousKeyBFE` is not defined here. I'll push up a counter PR with some minor fixes to this file so please wait for that :)

feed-formats/buttwoo.js
```diff
@@ -0,0 +1,361 @@
...
+
+const BUTTWOO_FEED_TF = bfe.toTF('feed', 'buttwoo-v1')
+
+module.exports = function init(ssb) {
+  const feedFormat = {
+    name: 'buttwoo',
+    encodings: ['js', 'bipf'],
+
+    getFeedId(nativeMsg) {
+      const pAuthor = 6 // we can assume this because the feed format is fixed
```
arj03, line 24: this doesn't work because length is stored as a varint, so for larger messages this is not 6 ;-)

feed-formats/buttwoo.js
```diff
@@ -0,0 +1,361 @@
...
+function base64ToBuffer(str) {
+  var i = str.indexOf('.')
+  return Buffer.from(str.substring(0, i), 'base64')
+}
+
+const BUTTWOO_FEED_TF = bfe.toTF('feed', 'buttwoo-v1')
+
+module.exports = function init(ssb) {
+  const feedFormat = {
+    name: 'buttwoo',
```
arj03, line 20: could we call this buttwoo-v1? That would make it consistent everywhere.

feed-formats/buttwoo.js
staltz: Damn. I thought I had a smart idea here :smiling_face_with_tear:

feed-formats/buttwoo.js
staltz: I fixed this in a recent commit, it should be orthogonal to https://github.com/ssbc/ssb-db2/pull/348

staltz:
Also, https://github.com/ssbc/ssb-db2/pull/347/commits/547df2537feea39c56b5a6e03617f7898dbebfee may look interesting, it creates the files

- `decrypted.index` (former canDecrypt.index)
- `encrypted-box1.index` (new)
- `encrypted-box2.index` (former encrypted.index)

mixmix:
Nice work btw

# :horse_racing: :lizard: :melon:

core.js
```diff
@@ -372,114 +412,324 @@ exports.init = function (sbot, config) {
...
+        feedFormat.validateSingle(hmacKey, nativeMsg, prevNativeMsg, (err) => {
           // prettier-ignore
-          if (err) return cb(clarify(err, 'bendy butt message validation in addImmediately() failed'))
-          const key = bendyButt.hash(msgVal)
-          updateState({ key, value: msgVal })
-          log.add(key, msgVal, (err, kvt) => {
+          if (err) return cb(clarify(err, 'addImmediately() failed validation for feed format ' + feedFormat.name))
+          const msgId = feedFormat.getMsgId(nativeMsg)
+          const msgVal = feedFormat.fromNativeMsg(nativeMsg, opts.encoding)
+          state.update(feedId, nativeMsg)
```
arj03, line 444: before this was a kvt, and for classic this is fine, we have the key there. But for nativeMsg we don't. So we need a way to store the key here as well. Will speed up validate.

compat/publish.js
```diff
@@ -0,0 +1,82 @@
...
+          'This is to protect your feed from forking ' +
+          'into an irrecoverable state.'
+      )
+    }
+  }
+
+  function encryptContent(keys, content) {
+    if (
+      sbot.box2 &&
+      sbot.box2.supportsBox2 &&
```
arj03, line 28: this doesn't seemt to be defined?

core.js
```diff
@@ -89,21 +81,23 @@ exports.init = function (sbot, config) {
...
   const indexes = {}
   const dir = config.path
   const privateIndex = PrivateIndex(dir, sbot, config)
-  const log = Log(dir, config, privateIndex)
+  const log = Log(dir, config, privateIndex, self)
   const jitdb = JITDB(log, jitIndexesPath(dir))
   const status = Status(log, jitdb)
   const debug = Debug('ssb:db2')
-  const post = Obv()
+  const post = Obv() // FIXME: move to compat?
```
arj03, line 93: Yes pleased! This is used in ebt, but that will change with https://github.com/ssbc/ssb-ebt/pull/68 anyway. This seems to be only used in migrate inside db2. Can you check manyverse if there are other places?

core.js
```diff
@@ -134,42 +128,70 @@ exports.init = function (sbot, config) {
     })
   })
 
+  // FIXME: we need to wait for box1 and box2 async "ready" to load
```
arj03, line 131: Yes

compat/publish.js
staltz: Yeah it's not, but I'll come back to this file later. I want the implementation to just use ssb.db.create

debounce-batch.js
```diff
@@ -19,21 +19,21 @@ module.exports = class DebouncingBatchAdd {
     this.queueByAuthor.delete(authorId)
     this.timestampsByAuthor.delete(authorId)
     // Add the messages in the queue
-    this.addBatch(msgVals, (err, kvts) => {
+    this.addBatch(msgVals, queue[0][1], (err, kvts) => {
```
arj03, line 22: could we do something like const opts = queue[0][1]. And then pass that in here. Would make things easier to read.

test/basic.js
```diff
@@ -369,11 +368,7 @@ test('add transaction ooo', (t) => {
 })
 
 test('add some bendybutt-v1 messages', (t) => {
-  const mfKeys = ssbKeys.generate()
-  const classicUri = SSBURI.fromFeedSigil(mfKeys.id)
-  const { type, /* format, */ data } = SSBURI.decompose(classicUri)
-  const bendybuttUri = SSBURI.compose({ type, format: 'bendybutt-v1', data })
-  mfKeys.id = bendybuttUri
+  const mfKeys = ssbKeys.generate(null, null, 'bendybutt-v1')
```
arj03, line 371: 🚀

arj03:
This is quite a beast, but overall a really nice and needed refactor :-) I have gone through a first pass of the code now.

Some overall comments:

> compat/publish.js

This is really nice!

> db.js → core.js

Much better name, and really like how this is now just the core stuff.

> addGroupKey

We should probably discuss how we define these groups sooner rather than later.

staltz:
> db.js → core.js

@arj03 My idea with the rename is that you could use ssb-db2 without anything extra, i.e. `.use(require('ssb-db2/core'))` such that you add your own feed formats and plugins, etc. `.use(require('ssb-db2/db'))` wouldn't be as nice. I considered 'bare.js' as a name for a while, but went with core which is more common for us. If you do `.use(require('ssb-db2'))` then you're using reasonable defaults and built-in stuff, although opinionated.

core.js
```diff
@@ -372,114 +412,324 @@ exports.init = function (sbot, config) {
...
+        })
+      }
+    )
+  }
+
+  function create(opts, cb) {
+    const guard = guardAgainstDuplicateLogs('create()')
+    if (guard) return cb(guard)
+
+    const keys = opts.keys || config.keys
```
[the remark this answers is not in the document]
staltz: > Maybe require keys to be defined if format is not classic?

Yes, we can do that, but additionally we can add a check: `keys.id` has to pass `feedFormat.isAuthor(keys.id)`, otherwise an error is thrown. That's because we now have `ssbKeys.generate(null, null, 'bendybutt-v1')`, we can assume that the resulting `keys.id` is correctly a bendybutt SSB URI.

arj03:
Yeah that is really nice. At some we can move out migrate as well, that will make package.json a lot leaner :)

core.js
staltz: That's a good idea! Some back story to how it ended up like this: buttwoo validation has `previousData`, and that's the array of arrays thing, not the previous KVT. So this led to me having to convert `previousKVT` (in the `state`) to a "nativeMsg" (a `previousData` for buttwoo). I then removed the conversion, so that `state` contains the `nativeMsg` directly, and passes that unconverted to the validation layer.

But indeed, I can add the msgId here.

debounce-batch.js
staltz: done

core.js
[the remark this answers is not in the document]
staltz: done

staltz:
Beautiful :partying_face: https://github.com/ssbc/ssb-db2/pull/347/commits/f68c43bd51061d447a5f491cdbae14a2e6a3efe6

core.js
staltz: In Manyverse, `post` is only used in `ssb-ebt` and `ssb-conn/scheduler`. Can easily be replaced

mixmix:
@staltz re "getRecipients" do you mean it to be "getEnctptionKeys"?

You didn't say what the buffers were

core.js
arj03: Seems to be mostly fixed by weakmap caching, at least for buttwoo

debounce-batch.js
```diff
@@ -63,10 +64,10 @@ module.exports = class DebouncingBatchAdd {
     }, this.period * 0.5)
   }
 
-  add(msgVal, cb) {
+  add(msgVal, opts, cb) {
```
arj03, line 67: this function needs to be aware of buttwoo's author + parent type

arj03: You need to bucket them into correct chains (so previous matches) and right now we don't, meaning that if you have parent on a message then it will be put into the same bucket as one without and validation will fail with wrong previous. There is a test in https://github.com/ssbc/ssb-ebt/pull/68 that fails because of this :)

staltz: Ooh, interesting

staltz: I can push a commit soon that fixes this.

staltz:
@arj03  I replaced `encryption-formats/box2` with `ssb-box2` module and tests failed for a reason that is very anticlimax:

```
Uncaught Exception at  uncaughtException : Error [OpenError]: IO error: lock /tmp/ssb-db2-about-self-index/keyring/LOCK: already held by process
```

This means that the `level` instance in `ssb-keyring` needs to be closed when the sbot is closed. =P

This requires some boring plumbing:

- encryptionFormat "spec" can have `teardown` method
- ssb-keyring needs to be updated with a close API
- ssb-box2 needs to have a teardown method that calls ssb-keyring close
- ssb-db2 needs to call teardown for all encryption formats, when sbot is closing

To be done.

staltz:
Okay, this branch is ALMOST done. `encryption-formats/box2` was replaced with `ssb-box2` module, and encryption formats `teardown()` is called on sbot.close. All tests pass *except* a simple one, pending on https://github.com/ssbc/ssb-private-group-keys/pull/14 being fixed and published. Before we publish the next major version of ssb-db2 with formats, it would be good that `ssb-keyring` depends on a published version of `ssb-private-group-keys` not on the PR branch.

staltz:
I'm gonna fix the benchmarks before this is ready for review.

arj03:
Some review notes:
 - the opts that `add`, `addOOO`, `addOOOBatch`, `addBatch`, `addTransaction` now can take needs to be documented
 - Can we make `add()` and friends default to encoding = js like create does?

Do you remember why we pin ssb-friends to 5.1.0?

arj03:
Looking at create and the `encryptionFormat` option. If you don't specify encryptionFormat, the function will just try each format to find one that works. Maybe we should default to `box` instead?

arj03:
Overall I think this looks really good. I love how core.js is now really streamlined. It's hard to review such a big PR, not really a fault, just mentioning it to say that we might find bugs when we start upgrading other modules to work with this. I'm happy that we have a relatively good test-suite (and benchmarks :)). So in order to get move things along and really test I'll say it's fine to merge this in once the minor things I mentioned are fixed together with the benchmarks as you mentioned. It is after all a major version change.

staltz:
Thanks @arj03 for taking a look, I'm working on those.

Two notes:

- I *disabled* the benchmark that uses box2 groups simply because we still don't have the removeGroupKeys API in ssb-box2 :(
- In `index.js`, I put both `ssb-box` and `ssb-box2`, do you think it that `ssb-box2` should not included (and the user has to install the plugin themselves)? Previously, ssb-db2 didn't support box2 out of the box, you had to install ssb-db2-box2

arj03:
I think it's good to include both box1 and 2 in index.js. just that the default would be box1 of you don't specify in create. It is still the case that most clients can't decide box2.

staltz:
>     * the opts that `add`, `addOOO`, `addOOOBatch`, `addBatch`, `addTransaction` now can take needs to be documented

Yes, done

>     * Can we make `add()` and friends default to encoding = js like create does?

Yes, done

> Do you remember why we pin ssb-friends to 5.1.0?

Oh, that's just a devDep and it's only used for the benchmarks (and actually those specific benchmarks are `skip`ped at the moment). It's pinned so that the benchmarks are always using the same ssb-friends and we would just be testing changes in ssb-db2 (not, for instance, performance improvements in ssb-friends).

staltz:
> Looking at create and the `encryptionFormat` option. If you don't specify encryptionFormat, the function will just try each format to find one that works. Maybe we should default to `box` instead?

This actually made a lot of sense, and I made the change. But then some tests broke. Specifically, if you use `ssb.db.publish()` with some private group as a recp, you don't have any way of specifying which encryptionFormat to use, so it'll use `box` and that's going to fail.

I'm inclined to use `ssb.db.create()` instead, but I wanted to let you know that if we do the change you suggested, you won't be able to use box2 with `ssb.db.publish()`.

arj03:
I think that is for the best actually as publish is legacy anyway

staltz:
Okay but I implemented it anyway. Ready for Final FINAL review.

arj03:
I think it will be for the best this way :) Nice to see the last 2 commits actually removed a bunch of lines

arj03:
Can you create an issue for the benchmark that was commented out?

arj03 (approved):
Great work on this. Super excited about the new ability to more easily add and structure feed formats :+1: :star: :1st_place_medal:

staltz:
MERGEDDDDDDD

PR #348, opened 2022-06-08
Minor fix to hash validation

arj03:
Hashes are encoded, so we get some agility if we want to support newer hash functions later.

We need to figure out how we merge what you have here in feed-formats and the ssb-buttwoo module, but for now I wanted this to work so I could test ebt :)


feed-formats/buttwoo.js
```diff
@@ -15,6 +15,13 @@ function base64ToBuffer(str) {
 
 const BUTTWOO_FEED_TF = bfe.toTF('feed', 'buttwoo-v1')
 
+function hashToBFE(buffer) {
+  return Buffer.concat([
+    Buffer.from([0]),
+    buffer
+  ])
+}
```
staltz, line 23: I think the reason why I removed this is that I didn't understand this. It's not really BFE, is it? It's missing the `F` byte, and the `T` byte is zero, which means "feed". Why do we need this, and couldn't we add new hash functions as `buttwoo-v2`, `buttwoo-v3` etc?

arj03: We could, but added buttwoo-v2 support requires changes something like 10 libraries. But yes, the name is super bad, I'll that.

arj03: And that is only on the JS side :)

arj03:
Tidied this, should be ready now

PR #352, opened 2022-06-12
Faster single add

arj03:
`add()` is used in EBT so should be as fast as possible, this avoids some overhead. classic add goes from 8400 to 8800 with this change.

### DIFF
--- a/core.js
+++ b/core.js
@@ -349,7 +349,7 @@ exports.init = function (sbot, config) {
         }
         opts.feedFormat = feedFormat.name
         if (feedFormat.validateBatch) debouncer.add(nativeMsg, opts, cb)
-        else addSingleMessageHelper(nativeMsg, feedFormat, opts, cb)
+        else addImmediately(nativeMsg, feedFormat, opts, cb)
       }
     )
   }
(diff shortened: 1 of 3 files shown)


core.js
```diff
@@ -413,11 +413,35 @@ exports.init = function (sbot, config) {
     )
   }
 
+  function addSingleMessageHelper(nativeMsg, feedFormat, opts, cb) {
```
staltz, line 416: Ah, I see. We can drop the internal checks.

Actually, `addImmediately` is an internal method, so we can just delete `addImmediately` and rename `addSingleMessageHelper` to be the new `addImmediately`.

arj03: Oh, it's [exposed[(https://github.com/ssbc/ssb-db2/blob/b1de5a4e57fd2255e8011a011fc11130fa5d0a86/core.js#L1101), just not in the manifest

staltz: Does anyone use it? Tests don't use it. It's exposed under the section "used internally by other plugins in this project (ssb-db2)".

arj03: Hmm, probably not. I'll fix that

arj03:
Should be ready now :)

staltz:
Awesome

arj03:
I think we can start looking at the uri changes e.g. now and see if we can add some of the things back

staltz:
URI changes, you mean the buttwoo URI with author+parent?

Add which things back?

arj03:
Things like this https://github.com/ssbc/ssb-db2/pull/347/files#diff-decb92c24aa3521a8f55136e6d18b8b5e8ca42fec174f0882079bf39f6a4e265R98

staltz:
Hmm, the link just points me to the whole PR

PR #353, opened 2022-06-12
Faster buttwoo validate

arj03:
Before: 494.17ms
After: 479.41ms

Biggest problem was that previous MsgIdBFE was not cached, so that had to be recalculated

### DIFF
--- a/feed-formats/buttwoo-v1.js
+++ b/feed-formats/buttwoo-v1.js
@@ -390,7 +390,7 @@ module.exports = function init(ssb) {
         )
 
       // Fast:
-      const public = authorBFE.slice(2).toString('base64')
+      const public = authorBFE.slice(2)
       // Proper:
       // const { data: public } = SSBURI.decompose(bfe.decode(authorBFE))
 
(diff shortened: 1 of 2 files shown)


arj03:
With this, EBT is now at 7265. Old was around 7500 so I think we can consider this good enough :)

feed-formats/buttwoo-v1.js
```diff
@@ -92,14 +98,25 @@ module.exports = function init(ssb) {
...
+      if (feedFormat._msgIdBFECache.has(nativeMsg)) {
+        return feedFormat._msgIdBFECache.get(nativeMsg)
+      }
+
+      let data = feedFormat._msgIdCache.get(nativeMsg)
+      if (!data) {
+        const [encodedValue, signature] = feedFormat._extract(nativeMsg)
+        data = blake3.hash(Buffer.concat([encodedValue, signature]))
+        feedFormat._msgIdCache.set(nativeMsg, data)
+      }
```
staltz, line 115: Nice optimization! Perhaps these 5 lines above can be refactored into a `_getMsgIdBuf` method that both `getMsgId` and `_getMsgIdBFE` can make use of.

Issue #354, opened 2022-06-22
Reset status and progress after compaction is done


staltz:
`Status` is stateful (particularly with leveldb indexes) and may make the calculation of progress bars become incorrect. We should reset it.

PR #356, opened 2022-06-22
status() is reset after compaction

staltz:
## Context

I'm putting compaction in Manyverse and realized that the progress bars were quite funky (going backwards sometimes).

## Problem

#354 

> `Status` is stateful (particularly with leveldb indexes) and may make the calculation of progress bars become incorrect. We should reset it.

## Solution

Add a new method `reset()` to `status` which puts the numbers as they were at initialization.


arj03 (approved):
👍

PR #361, opened 2022-06-29
fix migrate: close newLog when there  is no ssb-db2 installed

staltz:
## Context

I was testing Manyverse, and noticed that right after it migrates `~/.ssb` to ssb-db2 (using `ssb-db2/migrate` *without* the rest of ssb-db2), indexing would not happen at all, the app seemed totally blank without content. On a second run of the app, indexing happens and content appears.

## Problem

When using the migrate plugin alone, without ssb-db2, then the `sbot.close` hook does **not** run `log.close` (which should make sure all writes are done). This meant that `sbot.close` callbacks were called *before* AAOL was reading writing to disk.

## Solution

ssb-db2/db.js already handles this case correctly, so we only need to patch the case where we are running `.use(require('ssb-db2/migrate'))` without `.use(require('ssb-db'))`, such that the close hook calls `log.close`.

### DIFF
--- a/migrate.js
+++ b/migrate.js
@@ -208,6 +208,7 @@ exports.init = function init(sbot, config) {
   function oldLogMissingThenRetry(retryFn) {
     if (!hasCloseHook) {
       sbot.close.hook(function (fn, args) {
+        stop()
         if (!sbot.db && newLog) {
           newLog.close(() => {
             fn.apply(this, args)
--- a/migrate.js
+++ b/migrate.js
@@ -200,6 +200,7 @@ exports.init = function init(sbot, config) {
   const progressStream = Notify()
   let started = false
   let hasCloseHook = false
+  let newLog
   let retryPeriod = 250
   let drainAborter = null
   let liveProgressInterval = null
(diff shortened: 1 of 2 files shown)


arj03 (approved):
Change simple enough. Good find :)

arj03:
Oh, benchmark found a problem :)

arj03:
I didn't see that stop was an exposed function

staltz:
@arj03 Well spotted. I was lazy to change stop() to async (and make a breaking change version) so I went with a simpler alternative instead, does almost the same.

arj03:
So you don't need all what stop() does at this place? Like drainAborter.abort etc.? I'm not super familiar with this code.

staltz:
Well, things were working before without calling stop(), but I guess I can add `stop()` to the close hook.

It's a shame that flumelog doesn't have a `close()`, otherwise I'd call that too, on stop().

arj03 (approved):
Alright, 👍  assuming the tests and benchmark passes now :)

PR #362, opened 2022-06-29
do not allow concurrent updateIndexes() calls

staltz:
## Context

I was testing Manyverse and noticed with `DEBUG=ssb:db2` that there were really two different `updateIndexes()` calls running at startup (in normal use case circumstances).

## Problem

`ReadyGate.onReady` adds the cb to an array, **even if that cb was already added** to the array. So two different calls of `onReady` are *not* idempotent, causing the `cb` to be run twice when the ReadyGate is ready.

## Solution

Use a Set of functions instead of an Array of functions, inside ReadyGate. 

I also added a `throw` in updateIndexes() in case there is another updateIndexes() ongoing. The purpose of this was to fail the tests, but strangely, the process exit code is green, even though the test suite "failed". Weird, and searching for `unhandledRejection` in `tape` issues left me confused.

### DIFF
--- a/db.js
+++ b/db.js
@@ -570,7 +570,10 @@ exports.init = function (sbot, config) {
 
   function updateIndexes() {
     if (!log.compactionProgress.value.done) return
-    if (abortLogStreamForIndexes) throw Error('updateIndexes() already running')
+    if (abortLogStreamForIndexes) {
+      debug('updateIndexes() called while another one is in progress')
+      return
+    }
     const start = Date.now()
 
     const indexesArr = Object.values(indexes)


db.js
```diff
@@ -570,6 +570,7 @@ exports.init = function (sbot, config) {
 
   function updateIndexes() {
     if (!log.compactionProgress.value.done) return
+    if (abortLogStreamForIndexes) throw Error('updateIndexes() already running')
```
arj03, line 573: I think I would prefer to write this to log and return instead of throwing an error, because nothing is catching those errors so it would be a hard crash.

arj03:
The set change looks great :)

arj03 (approved):
Very nice. Some really good fixes you are finding here.

PR #363, opened 2022-07-01
implement alternative to level.clear()

staltz:
## Context

Post-compaction reindexing was almost 2x slower than initial indexing, even though it ran on less msgs. See https://github.com/ssbc/ssb-db2/issues/306#issuecomment-1171672278

## Problem

Leveldown clear() (used when resetting a leveldb index) is notoriously slow, because it does something naive: it iterates over each db item (in JS!) and deletes one *at a time*. Not even parallelism is employed.

## Solution

We implement clear() ourselves with a nuclear approach: delete the folder and recreate level.

### DIFF
--- a/indexes/plugin.js
+++ b/indexes/plugin.js
@@ -12,6 +12,14 @@ const Debug = require('debug')
 const DeferredPromise = require('p-defer')
 const { indexesPath } = require('../defaults')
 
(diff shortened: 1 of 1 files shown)


arj03 (approved):
nice and simple 👍  (the solution, not finding it ;-))

PR #365, opened 2022-07-07
improve deleteFeed() idempotence performance

staltz:
**Context:** It would be good if ssb-db2 users could just call `deleteFeed()` even for feeds that are already deleted (thus "idempotent" operation) and not have to worry about that being a slow operation.

I just made a simple change where `deleteFeed()` checks if we even have any message for such feed, and if we don't, immediately respond the callback.

(I can rebase `formats-split` on this after we merge this)


arj03 (approved):
Yes

PR #367, opened 2022-07-13
dont allow queries to start when compact() just started

staltz:
## Problem

`compacting` is set to true only when `compactionProgress` reports something, but that could take a while (it needs to write a block in AAOL), and in the meantime, someone could call `ssb.db.query`, which can proceed if `compacting === false`.

## Solution

As soon as we call `compact()`, set `compacting` to true.

### DIFF
--- a/db.js
+++ b/db.js
@@ -816,6 +816,7 @@ exports.init = function (sbot, config) {
     fs.closeSync(fs.openSync(resetLevelPath(dir), 'w'))
     fs.closeSync(fs.openSync(resetPrivatePath(dir), 'w'))
     fs.closeSync(fs.openSync(reindexJitPath(dir), 'w'))
+    compacting.set(true)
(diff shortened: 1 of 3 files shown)


db.js
```diff
@@ -816,6 +816,7 @@ exports.init = function (sbot, config) {
     fs.closeSync(fs.openSync(resetLevelPath(dir), 'w'))
     fs.closeSync(fs.openSync(resetPrivatePath(dir), 'w'))
     fs.closeSync(fs.openSync(reindexJitPath(dir), 'w'))
+    compacting.set(true)
```
arj03, line 819: Wouldn't it be better to move this up 3 lines, du before the deletes?

arj03: Nevermind

arj03 (approved):
👍

PR #368, opened 2022-07-16
Use unbox function

arj03:
This is mostly a code cleanup by using the unbox function since we don't need the key here. There is also a tiny performance improvement because the derived secret is only calculated [once](https://github.com/ssbc/envelope-js/blob/8532da86ea84f79ae2532b17e2768586e1c40da6/unbox.js#L17)


staltz (approved):
All good in my opinion. But you could also mirror these changes on the ssb-db2-box2 PR because once that one is done, we'll delete this file in db2

PR #369, opened 2022-07-16
Improve box1 & box2 benchmarks

### LINKED ISSUE #346: Optimize box2 decryption for multiple recps
We need to figure out why box2 decryption is around 3 times slower than box1. See https://github.com/ssbc/ssb-db2/pull/342#issuecomment-1125980482

arj03:
Right now the box1 and box2 benchmarks don't decrypt anything because no messages are for you. While in the groups benchmark 30% is for you. This is confusion when reasoning about performance and also not super realistic. This PR changes things so that we can more easily compare things. See https://github.com/ssbc/ssb-db2/issues/346 for some numbers.

With this I think we can close #346. The conclusion is in https://github.com/ssbc/ssb-db2/issues/346#issuecomment-1184148122.

### DIFF
--- a/benchmark/index.js
+++ b/benchmark/index.js
@@ -287,10 +287,18 @@ test('box1', (t) => {
     .call(null, { keys: keys2, path: dirBox1 })
 
   const recps = [...randos, keys.id]
+  const recpsMe = [...randos, keys2.id]
+
+  // for messages, the percentage of groups that includes you
+  const percentMe = 30
 
   let contents = []
-  for (var i = 0; i < 1000; ++i)
-    contents.push({ type: 'tick', count: i, recps: shuffle(recps) })
+  for (var i = 0; i < 1000; ++i) {
+    const percentage = Math.floor(Math.random() * 100)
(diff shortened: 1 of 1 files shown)


benchmark/index.js
```diff
@@ -287,10 +287,18 @@ test('box1', (t) => {
...
+
+  // for messages, the percentage of groups that includes you
+  const percentMe = 30
 
   let contents = []
-  for (var i = 0; i < 1000; ++i)
-    contents.push({ type: 'tick', count: i, recps: shuffle(recps) })
+  for (var i = 0; i < 1000; ++i) {
+    const percentage = Math.floor(Math.random() * 100)
+    const msgRecps = percentage < percentMe ? shuffle(recps) : shuffle(recpsMe)
```
staltz, line 298: Should the ternary be flipped around?

arj03: :facepalm: yes it should. Nice catch

staltz (approved):
Yup

PR #370, opened 2022-07-18
use jitdb@7.0.0

staltz:
Remove `post-compact-reindex-jit` state file from ssb-db2 and use what jitdb 7.0.0 has instead.

### DIFF
--- a/compat/history-stream.js
+++ b/compat/history-stream.js
@@ -3,10 +3,9 @@
 // SPDX-License-Identifier: LGPL-3.0-only
 
 const pull = require('pull-stream')
-const pullCont = require('pull-cont')
 const ref = require('ssb-ref')
 const Hookable = require('hoox')
-const { author } = require('../operators')
+const { author, and, gte, where, batch, toPullStream } = require('../operators')
 const { reEncrypt } = require('../indexes/private')
 
 // exports.name is blank to merge into global namespace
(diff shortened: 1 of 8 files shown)


compat/history-stream.js
```diff
@@ -21,28 +20,18 @@ exports.permissions = {
 
 exports.init = function (sbot, config) {
   sbot.createHistoryStream = Hookable(function createHistoryStream(opts) {
```
staltz, line 22: I had to reimplement createHistoryStream using `toPullStream()` instead of `jitdb.paginate` because the former keeps track of `latestMsgPrecompaction` while we would have to manually do it with the latter.

db.js
```diff
@@ -547,27 +546,24 @@ exports.init = function (sbot, config) {
...
   }
 
-  function restartUpdateIndexes() {
-    if (abortLogStreamForIndexes) {
-      abortLogStreamForIndexes()
-      abortLogStreamForIndexes = null
+  function resumeUpdatingIndexes() {
+    if (abortLogStreamForIndexes) return
+    else indexesStateLoaded.onReady(updateIndexes)
+  }
```
staltz, line 559: These functions are already in the `formats-split` branch but I had to backport them here.

test/compaction.js
```diff
@@ -70,7 +70,7 @@ test('compaction fills holes and reindexes', async (t) => {
     done = true
   })
 
-  const expectedProgress = [0.0008620390691959752, 1]
+  const expectedProgress = [0, 0.0008620390691959752, 1]
```
staltz, line 73: These changes are because of the newer version of AAOL

arj03 (approved):
:+1: lets hope this is the last one :)

staltz:
Yes! Actually, I put this branch in manyverse and so far no crashes or bugs :)

I'll still test it in all possible brutal ways, but I do have hope this is good now.

arj03:
image

PR #371, opened 2022-07-20
wait for encryption formats to load before encrypting

staltz:
**Problem:** our `./encryption-formats/box2.js` `setup()` was actually not asynchronous, it was sync. Because of this, you could always call any of its APIs at any time, and it would work. Add a simple `setTimeout(1000` and then it doesn't work anymore.

The deeper problem is that in `indexes/private.js` we load the index files and if there is an error (and there often is, because the index files are non-existent) then the *other* computation (such as loading encryption formats) is cancelled. This is a normal behavior of errors in JS, and `multicb` does it too: if there is an error anywhere, stop everything.

But what we really want is: if there is an error when loading private indexes, that's okay, please proceed async loading the encryption formats too.

**Solution:** in `indexes/private`, treat loadFile errors as "callback results", not as errors, and treat encryptionFormats.setup errors as critical (which they are). Simulate async in `encryption-formats/box2.js`, and then also make sure that `create()` waits for private indexes (and thus encryption formats) to load before proceeding to encrypt.

### DIFF
--- a/core.js
+++ b/core.js
@@ -26,7 +26,7 @@ const {
   resetLevelPath,
   resetPrivatePath,
 } = require('./defaults')
-const { onceWhen, ReadyGate } = require('./utils')
+const { onceWhen, ReadyGate, onceWhenPromise } = require('./utils')
 const DebouncingBatchAdd = require('./debounce-batch')
 const Log = require('./log')
 const Status = require('./status')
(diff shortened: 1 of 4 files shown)


core.js
```diff
@@ -620,7 +620,7 @@ exports.init = function (sbot, config) {
     )
   }
 
-  function create(opts, cb) {
+  async function create(opts, cb) {
```
arj03, line 623: Since create is not called very often it's probably fine to do this one as async, I just prefer to have them only in tests normally.

arj03: Yeah thought that was the case, just beware of the dreaded promises :dragon_face:

PR #372, opened 2022-07-21
benchmark deleteFeed()

staltz:
I thought I was going to speed up `deleteFeed()` but I wanted to first make a benchmark to prove it's getting faster. Then I tried all sort of things and I realized I could only make `deleteFeed()` *slower* :cry: . But at least it's useful to add this benchmark so we know the baseline for deleteFeed.


arj03 (approved):
:+1:

Issue #374, opened 2022-07-22
Bring back isPrivate() operator


staltz:
With the semantics `or(isEncrypted(), isDecrypted())`, because why not.

PR #375, opened 2022-07-23
bring back isPrivate() operator

staltz:
For issue #374. I noticed that I would just use `or(isEncrypted(), isDecrypted())` and that's a bit verbose so we can just have a shortcut.

See https://github.com/ssbc/ssb-subset-ql/blob/1f99976f59fc105b15bbab6fce4802d6d585580e/ql0.js#L89


arj03 (approved):
Yes

PR #376, opened 2022-07-27
use monotonic-timestamp in create()

staltz:
`ssb-db` and `ssb-validate` both use `monotonic-timestamp` to avoid the problem where two consecutive calls of `Date.now()` would give the same number. In our case this would help so we can avoid the `Date.now() + 1`, `Date.now() + 2` hacks. I do think that the async-await we use in `create()` already guarantees that `Date.now()` will be different (I was unable to create a 1st :x: 2nd :heavy_check_mark: test), but it makes sense to anyway use `monotonic-timestamp` to be sure we never bump into this problem.

### DIFF
--- a/log.js
+++ b/log.js
@@ -5,6 +5,7 @@
 const OffsetLog = require('async-append-only-log')
 const bipf = require('bipf')
 const TooHot = require('too-hot')
+const newTimestamp = require('monotonic-timestamp')
 const { BLOCK_SIZE, newLogPath, tooHotOpts } = require('./defaults')
 
 const BIPF_AUTHOR = bipf.allocAndEncode('author')
@@ -37,7 +38,7 @@ module.exports = function (dir, config, privateIndex, db) {
     const kvt = {
       key,
       value,
-      timestamp: Date.now(),
+      timestamp: newTimestamp(),
     }
     if (feedId !== value.author) kvt.feed = feedId
     const recBuffer = bipf.allocAndEncode(kvt)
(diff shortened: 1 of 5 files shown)


log.js
```diff
@@ -62,11 +63,12 @@ module.exports = function (dir, config, privateIndex, db) {
     let recBuffers = []
     let kvts = []
 
+    const timestamp = newTimestamp()
```
staltz, line 66: This means that all msgs in the transaction would have the same timestamp. Is that okay/expected?

staltz:
I'm not sure what's going on with the tests, they pass locally, but we definitely need to fix them in CI before merging.

log.js
arj03: They need to be unique. Right now we only use this function for indexes where the 2 mesages are on different feeds, but in general if you add 2 messages to the same feed, then validate will complain.

staltz:
The reason why tests were broken was that `monotonic-timestamp` was being replaced by `mock-monotonic-timestamp` (because the way we use ssb-fixtures ends up "leaking" this hack globally) so that there was a test that checked that the `msg.timestamp` should be greater than some Date.now(), but the `msg.timestamp` was in the year 2015, because of `mock-monotonic-timestamp`.

I used `pnpm` in CI, which fixed that error, but then other problems came up: some timestamps are floats, like `1659105426666.001` which doesn't work in bendybutt because [ssb-bfe only bypasses integers](https://github.com/ssbc/ssb-bfe/blob/43c69822f01ff5898fd034776053dd1d832340dd/index.js#L137), but **also** it turns out that bencode only supports integers, not floats. The JS library `bencode` will convert the float to an integer.

@arj03 What do you think, should we quit this PR?

arj03:
Regarding bendy butt I guess it's a matter of interpretation. I really dislike that we allowed floating timestamps in classic and think it would be better overall to convert all of them to int.

staltz:
So does that mean we should make a new (version or fork) monotonic-timestamp that only gives integers?

arj03:
That would make this PR a lot more tasty :)

staltz:
Hmm, pretty sure we could rewrite a new PR instead of working on this old one. I want to kill some old PRs.

PR #377, opened 2022-08-01
add logStats() API

staltz:
This is just because I want to avoid `getLog()` as a public API (I'm trying to reduce the amount of APIs in each library, this gives us more implementation flexibility), but we need a way of calling the `stats` method in AAOL.


arj03 (approved):
👍 

It seems like the getLog api is only used in this module: https://github.com/search?q=org%3Assbc+getlog&type=code

PR #383, opened 2022-08-05
Throttle

arj03:
This takes care of https://github.com/ssbc/ssb-db2/issues/379

### DIFF
--- a/core.js
+++ b/core.js
@@ -27,7 +27,7 @@ const {
   resetPrivatePath,
 } = require('./defaults')
 const { onceWhen, ReadyGate, onceWhenPromise } = require('./utils')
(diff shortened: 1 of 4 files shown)


core.js
```diff
@@ -331,8 +331,8 @@ exports.init = function (sbot, config) {
     return null
   }
 
-  const debouncePeriod = config.db2.addDebounce || 250
```
arj03, line 334: note this was not matching README before ;-)

arj03:
I guess this is technically a major version bump

staltz (approved):
Looks good. Very simple change, indeed.

And yeah this would qualify as a breaking change. It's annoying, but let's not get lazy with these stuff. I'm fine with updating some modules to support db2 v6

PR #384, opened 2022-08-22
add reset() API

staltz:
We have this API hacked in Manyverse already, would be nice to upstream it in ssb-db2: https://github.com/staltz/manyverse/blob/b0f5f4b17e85ac6d2777bd581dd084fbf97f5b59/src/backend/patches/ssb-db2%2B4.2.1.patch


arj03 (approved):
Yep, looks correct

PR #385, opened 2022-08-22
add getIndexingActive() API

staltz:
We need this for pausing the ssb-friends / ssb-replication-scheduler integration, because indexingActive is triggered *right before* log.stream to update the indexes, and we need this "right before" behavior, that `status` doesn't give us.

I haven't thought out this API that well, I think in the future we could find a more developer friendly way of doing this.

### DIFF
--- a/core.js
+++ b/core.js
@@ -1090,6 +1090,7 @@ exports.init = function (sbot, config) {
     addOOO,
     addOOOBatch,
     getStatus: () => status.obv,
+    getIndexingActive: () => indexingActive,
     operators,
     onMsgAdded,
     compact,


arj03:
Well its nice that this an an obz so you can do something elsewhere depending on the status (changes) on this. Should we add these to the README (forgot to mention that with `reset`), or are they considered "internal" for now?

staltz:
> (forgot to mention that with `reset`)

Oh, I added README docs for `reset`, but not for `getIndexingActive`. I kind of not so proud about this API, I want to see if it's going to be actually handy for ssb-replication-scheduler.

arj03 (approved):
👍

PR #386, opened 2022-08-31
test streaming and decrypting records

staltz:
## Context

https://github.com/ssbc/ssb-replication-scheduler/issues/3

## Problem

In privateIndex we use `streaming === true` to update the latestOffset, but this is troublesome if there is some live stream which takes the latest record in the log, because it'll update privateIndex's latestOffset and then when reindexing all previous records, we won't attempt to decrypt them because they aren't in the decrypted list and neither does the `recOffset > latestOffset` pass.

## Solution

We should really only have one log.stream that is responsible for updating the private index. I elect the "leveldb log.stream" to be the only stream that can do that, thus we pass a `opts.shouldDecrypt = true` only for the core log.stream.

At the same time, this required fixing how log.stream was aborted. There was a race condition or something like that. We need to set the reference to log.stream.abort **before** we start draining the log.stream.

### DIFF
--- a/core.js
+++ b/core.js
@@ -787,7 +787,7 @@ exports.init = function (sbot, config) {
       debug('updateIndexes() called while another one is in progress')
       return
     }
-    const shouldDecrypt = true
+    const updatePrivateIndex = true
     const start = Date.now()
 
     const indexesArr = Object.values(indexes)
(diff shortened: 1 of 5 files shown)


staltz:
I broke tests elsewhere, but I'll check that tomorrow. @arj03 this is ready for a first-pass review (I have checked that the solution passes the test I added)

staltz:
@arj03 Okay, I think all tests pass now. What I did in the https://github.com/ssbc/ssb-db2/pull/386/commits/d7bc8644a13ce99a202d5db283f6af117dd6f778 commit was to allow jitdb log.streams to *decrypt* messages (and create the msg.meta), but **not** update the private index `latestOffset`. That's the sole responsibility of the core.js log.stream that updates leveldb indexes.

arj03 (approved):
Yes, this looks better overall that only 1 has the responsibility of updating the private index

PR #388, opened 2022-09-13
add publicWebHosting to aboutSelf

mixmix:
Problem:
- Planetary viewer system uses `ssb-db2` to serve threads, profiles over the web
- The current `indexes/about-self` don't track `publicWebHosting`

This PR:
- adds support for this to the aboutSelf index
- adds `isString` checks to `name` and `description` fields
    - @staltz in Mix's head "make it a different PR"
    - @mixmix "no, this has to be here, it's not safe without this check"
    - @arj03 "actually it's faster if you just chain heaps of `if &&`, what you've written is not easy for v8 to optimise"
    - @mixmix :sob:


staltz:
Hey @mixmix, some time ago I extracted this plugin out of ssb-db2 and put it at https://github.com/staltz/ssb-about-self. We're going to remove it from ssb-db2, so it's not worth updating it here. Can you make the PR over there?

mixmix:
Yeah sure thing. I'll likely make a pr which changes this repo to import that module and export it so there's no divergence of logic

indexes/about-self.js
```diff
@@ -2,107 +2,4 @@
...
-      live: true,
-      old: false,
-    })
-  }
-
-  getProfiles() {
-    return this.profiles
-  }
-}
+module.exports = require('ssb-about-self/plugin.js')
```
mixmix, line 5: @staltz now that I know about `ssb-about-self` I have scooped out the contents of this file, and replaced it with that modules logic so we have a single source of truth.

I've also killed the tests here, as that logic is now maintained elsewhere

about-self.js
```diff
@@ -2,8 +2,12 @@
 //
 // SPDX-License-Identifier: LGPL-3.0-only
 
+// DEPRECATED
+// please use https://github.com/ssbc/ssb-about-self
 const AboutSelfIndex = require('./indexes/about-self')
 
 exports.init = function (sbot, config) {
+  console.log('ssb-db2/about-self is deprecated, please use ssb-about-self')
```
staltz, line 10:
```suggestion
  console.warn('ssb-db2/about-self is deprecated, please use ssb-about-self')
```

README.md
```diff
@@ -181,30 +181,6 @@ sbot.db.query(
 )
 ```
 
-#### About-self
```
mixmix, line 184: I've removed this section of the README so people don't use this API here.
Alternatiively we could leave the heading and say:

```
There used to be a plugin hosted by this module. Please use `ssb-about-self` which now provides this functionality
```

mixmix:
@staltz can you review this again please?

README.md
staltz: Removing seems fine IMO

about-self.js
staltz: @mixmix

staltz: Check this @mixmix

staltz:
@mixmix I have, but you haven't apparently seen https://github.com/ssbc/ssb-db2/pull/388#discussion_r969434521

staltz:
Merged because you said in a private chat that 

> I would have preferred you just committed that, merged and published.

staltz:
v6.2.1

PR #389, opened 2022-09-14
allow heterogenous feed formats in addTransaction

staltz:
## Context

https://github.com/ssbc/ssb-ebt/pull/71

## Problem

We need to add an index message `[indexMsg, payload]` with the `sbot.db.addTransaction` API, such that the `indexMsg` is the "normal" part and the `payload` is the OOO part, but current addTransaction assumes that all of them have the same feed format.

## Solution

This PR changes addTransaction internally, in a backwards-compatible way, so that the "nativeMsgs" part can be a different feed format to the "oooNativeMsgs" part.

### DIFF
--- a/core.js
+++ b/core.js
@@ -536,23 +536,22 @@ exports.init = function (sbot, config) {
     if (nativeMsgs.length === 0 && oooNativeMsgs.length === 0) {
       return cb(null, [])
     }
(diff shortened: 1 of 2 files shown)


arj03 (approved):
Correct from what I can see 👍

PR #392, opened 2022-09-23
Improve documentation for getStatus

arj03:
@mixmix can you see if this helps you describe what the function does?

### DIFF
--- a/README.md
+++ b/README.md
@@ -554,7 +554,19 @@ also be read using `ssb.db.onMsgAdded.value`.
 
 ### getStatus
 
-Gets the current db status, either as a synchronous function or observeable
+Gets the current db status, either as a synchronous function or
+observeable. This is similar to
+[db.status](https://github.com/ssbc/ssb-db#dbstatus) in ssb-db. The
+value contains the following information:
+
+- `log`: current byte offset of the head of the log in bytes (the log
+  is the primary source of truth where raw messages are stored)
+- `jit`: an object with the offset of each [jitdb] index
+- `indexes`: an object with the offset of each level index
+- `progress`: a float representing the index completion from 0 to 1
+
+Example:
+
 ```js
 // synchronous
 const status = ssb.db2.getStatus()
@@ -565,13 +577,7 @@ const unsubscribe = ssb.db2.getStatus(listener)
 
 // later
 unsubscribe() // unsubscribes the listener from status updates
-[db.status](https://github.com/ssbc/ssb-db#dbstatus) in ssb-db. The
-value contains the following information:
-
-- `log`: current byte offset of the head of the log in bytes (the log is the primary source of truth where raw messages are stored)
-- `jit`: an object with the offset of each [jitdb] index
-- `indexes`: an object with the offset of each level index
-- `progress`: a float representing the index completion from 0 to 1
+```
 
 ### reindexEncrypted(cb)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -554,8 +554,14 @@ also be read using `ssb.db.onMsgAdded.value`.
 
 ### getStatus
 
-Gets the current db status, same functionality as
-[db.status](https://github.com/ssbc/ssb-db#dbstatus) in ssb-db.
+Gets the current db status, same functionality (observable) as
```
mixmix, line 557: I have only used it like
```
ssb.db2.getStartus()
```

does observeable mean you can go:

```js
const stop = ssb.db2.getStatus(status => {
    console.log(status)
})

stop()
```

arj03: yes :)

mixmix: I think saying you can use it both ways would be good

Example usages:

```js
const status = ssb.db2.getStatus()


const unsbscribe = ssb.db2.getStatus(status => console.log(status.progress)) 

// later
unsubscribe() // unsubscribes the listener from status updates
```

README.md
```diff
@@ -554,8 +554,14 @@ also be read using `ssb.db.onMsgAdded.value`.
...
+[db.status](https://github.com/ssbc/ssb-db#dbstatus) in ssb-db. The
+value contains the following information:
+
+- `log`: offset in log
```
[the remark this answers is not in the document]
arj03: This is the standard terminology we use in all db2 related modules: see https://github.com/ssbc/async-append-only-log/ README. 

```
This module is not compatible with flume without a wrapper around stream as it uses the same terminology as [JITDB](https://github.com/ssb-ngi-pointer/jitdb/) and [ssb-db2](https://github.com/ssb-ngi-pointer/ssb-db2/) of using **offset** for the byte position of a record instead of seq.
```

We made sure that this was consistent across all modules :-) Before these things could be very confusing with seq and sequence and what not :)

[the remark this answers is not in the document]
arj03: crap, format not working in comments...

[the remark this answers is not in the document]
mixmix: also what is the log?

```suggestion
- `log`: current offset of the head of the log in bytes (the log is the primary source of truth where raw messages are stored)
```

README.md
arj03: Well you can always call .value on an obz (example 1)

mixmix (changes requested):
It's an improvement, but it could still tell me a little more (as a beginner). Like the units would be the bare minimum.

Could also be worth saying something like:

> messages are written to the "log" (see `status.log` offset increasing), this triggers updating of the indexes (see `status.jit` for dynamic indexes and `status.indexes` for plugin defined indexes), which will have fallen behind. Once the indexes have completed indexing the new log messages, the offset numbers everywhere will be in sync, and you will see `status.progress = 1`

Another way to do this would be with an example

```js
{
    log: 2301,  // means we have 2301 bytes of messages in our primary database
    jit: {
        cat: 2301,  // this index is in sync (up to date) with the log
        dog: 1980,  // this index has ~300+ bytes still to index
    },
    progress: 0.87  //  this is <1 as jit.dog is still catching up 
}
```

mixmix:
@arj03 can we commit the suggestions I made, or some alternative. Anything small would be better than nothing merged

mixmix:
@arj03 polite bump. Lets go for easy improvements

arj03:
I'll have a look soon.

PR #393, opened 2022-10-03
No encryption

mixmix:
**Problem**: building a peer which doesn't do any encryption, but may want to publish things. Removing encryption reduces load on this peer (running on cheap droplet). The current code fails, even if you're not encrypting anything.

**Solution**: added a test case for minimal setup, and patched code to make it pass

### DIFF
--- a/test/minimal-db2.js
+++ b/test/minimal-db2.js
@@ -1,0 +1,35 @@
+const test = require('tape')
+const ssbKeys = require('ssb-keys')
+const rimraf = require('rimraf')
+const SecretStack = require('secret-stack')
+const caps = require('ssb-caps')
+
+test('minimal db2 (no encryption)', t => {
+  const dir = '/tmp/ssb-db2-minimal-create'
+  rimraf.sync(dir)
+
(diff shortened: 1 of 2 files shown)


test/minimal-db2.js
```diff
@@ -0,0 +1,42 @@
...
+  const keys = ssbKeys.generate()
+
+  const stack = SecretStack({ appKey: caps.shs })
+    .use(require('../core'))
+    .use(require('ssb-classic'))
+    // .use(require('ssb-box'))
+    // .use(require('ssb-box2'))
+    // .use(require('../compat/publish'))
+    // .use(require('../compat/post'))
+    // .use(require('../migrate'))
```
staltz, line 20: Can you remove this commented code? And add a license header for this file (see how CI checks are failing).

That's my only feedback, otherwise looks good to merge and makes sense.

staltz (dismissed):

mixmix:
That we commented choice I left there deliverately to show what had been
deactivated. But sure

PR #397, opened 2022-10-13
fix annoying encryptionFormat bug

mixmix:
I want to be able to just use box2, but when i took out box1 all my tests for encrypted messages were complaining.
ANNOYING AS SHIT

This inversion fixes it - i.e. only be fretting about encryptionFormat being present or not IF I'm possibly trying to encrrypt


staltz (approved):
Oh yeah, this makes absolute sense. Nice nano PR.

PR #398, opened 2022-10-13
Update ssb-box2

mixmix:
noticed the bundled dependency was outdated, so updated
Also updated other deps (minor versions)


package.json
```diff
@@ -43,33 +43,33 @@
     "pull-stream": "^3.6.14",
     "push-stream": "^11.0.0",
     "rimraf": "^3.0.2",
-    "ssb-about-self": "^1.0.1",
+    "ssb-about-self": "^1.1.0",
     "ssb-box": "^1.0.0",
-    "ssb-box2": "^2.0.2",
+    "ssb-box2": "^3.0.0",
```
mixmix, line 48: only *major* update

mixmix:
failing test, not sure why, don't have time today

package.json
staltz: I highly suspect the tests are breaking because of one of these deps. The easy way to fix the tests is to revert these changes and bump *only* ssb-box2. Unless you have the patience to debug which of these deps broke the tests and why...

arj03:
@mixmix

staltz:
I suppose we can just merge, because Mix has sick family at the moment.

mixmix:
yeah you should feel free to merge stuff like this

PR #400, opened 2022-10-19
private index updates the status

staltz:
## Context

https://github.com/ssbc/ssb-ebt/issues/77

When the private index *alone* was being rebuilt, no other index, then Manyverse would not show a progress bar.

## Problem

Private index is actually not included in the `status` object that is used for calculating indexing progress.

## Solution 

Now it is.

### DIFF
--- a/core.js
+++ b/core.js
@@ -118,6 +118,7 @@ exports.init = function (sbot, config) {
     })
   })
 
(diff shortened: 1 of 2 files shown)


arj03 (approved):
👍

PR #401, opened 2022-11-03
Accept URI IDs in get and getMsg

Powersource:
I'm trying to get a root msg in https://github.com/ssbc/ssb-tribes2/pull/29 but i have its id in uri form

### DIFF
--- a/core.js
+++ b/core.js
@@ -19,6 +19,7 @@ const Obv = require('obz')
 const bipf = require('bipf')
 const JITDB = require('jitdb')
 const jitdbOperators = require('jitdb/operators')
+const uri = require('ssb-uri2')
 
 const operators = require('./operators')
 const {
@@ -223,6 +224,8 @@ exports.init = function (sbot, config) {
   }
 
   function getHelper(id, onlyValue, cb) {
+    id = uri.isClassicMessageSSBURI(id) ? uri.toMessageSigil(id) : id
+
     if (sbot.db2migrate) {
       onceWhen(
         sbot.db2migrate.synchronized,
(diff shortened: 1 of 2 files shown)


staltz (approved):
Very decent PR, thanks.

Just as a general comment, I'm a bit unsure about this whole thing, just because ssb-db2 so far has been "feedFormat-agnostic", and all feed format logic is put into `ssb-classic`, `ssb-buttwoo` and `ssb-bendy-butt`. With this change, we have an exception here, where we add some classic-specific logic in the `getHelper()` function. This definitely increases the developer experience of using ssb-db2 (and it's why we should merge) but it opens up a new question:

Should we allow `ssb:message/classic/__` as a 1st class citizen *in the database*? I would like to think that `ssb-classic` is tied to `@__.ed25519` instead of having two representations. It makes life simpler. But there's also another option for the future, which is a migration path to "classic-v2" where URIs are used.

Maybe that's the underlying real concern: should "classic" be *only* sigils and should we make `classic-v2` that *only* has URIs? `classic-v2` is actually quite exciting to me, because it would not deviate a lot from classic-v1, it just tweaks a couple things to make it a better feed format. It would be fairly trivial for other implementations in Go and Python and what not to adapt to this. We could also fix a few other things in classic-v2, such as the V8 engine semantics for JSON.stringify.

Thoughts? @Powersource @arj03 (remember to ping arj in all things db2)

Also, feel free to press the big green.

Powersource:
Publish pls :)

Powersource:
I think/guess that just switching everythng to URIs at once is a simpler DX for developer users

PR #403, opened 2022-11-07
Simplify deleteFeed

### LINKED ISSUE #402: OOM when deleting huge feeds
My manyverse on desktop crashes after < 1 minute. I finally had some time to debug what is going on and found out that the problem is ssb-friends-purge. This module seems to try and delete around 1250 feeds when it starts up. One of the first feeds it finds is: @/iGURN4KIyJWMUDdbpjaZQHyCL9fjpFWGObGZ8zhzts=.ed25519. This feed has 121863 messages so when it tries to delete the feed, the asOffsets call [here](https://github.com/ssbc/ssb-db2/blob/master/core.js#L763) never finishes. It simply dies from out of memory. Is there any reason we can't just stream the offsets directly into log.del?

arj03:
This fixes #402. No need to load up all the offsets into memory if we just delete all of them anyway.

### DIFF
--- a/core.js
+++ b/core.js
@@ -757,24 +757,15 @@ exports.init = function (sbot, config) {
       () => {
         if (!state.has(feedId)) return cb()
 
-        self.query(
-          where(author(feedId)),
-          asOffsets(),
-          toCallback((err, offsets) => {
+        pull(
+          self.query(where(author(feedId)), asOffsets(), toPullStream()),
+          pull.asyncMap(log.del),
+          pull.collect((err) => {
             // prettier-ignore
-            if (err) return cb(clarify(err, 'deleteFeed() failed to query jitdb for ' + feedId))
+            if (err) return cb(clarify(err, 'deleteFeed() failed for feed ' + feedId))
 
-            push(
-              push.values(offsets),
-              push.asyncMap(log.del),
-              push.collect((err) => {
-                // prettier-ignore
-                if (err) return cb(clarify(err, 'deleteFeed() failed for feed ' + feedId))
-
-                state.delete(feedId)
-                indexes.base.removeFeedFromLatest(feedId, cb)
-              })
-            )
+            state.delete(feedId)
+            indexes.base.removeFeedFromLatest(feedId, cb)
           })
         )
       }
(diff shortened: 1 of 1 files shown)


arj03:
I can see that DeleteFeed at least here is much slower than before. I'll try giving it a look, but it is a bit strange.

core.js
```diff
@@ -757,24 +757,15 @@ exports.init = function (sbot, config) {
...
         if (!state.has(feedId)) return cb()
 
-        self.query(
-          where(author(feedId)),
-          asOffsets(),
-          toCallback((err, offsets) => {
+        pull(
+          self.query(where(author(feedId)), asOffsets(), toPullStream()),
+          pull.asyncMap(log.del),
+          pull.collect((err) => {
```
staltz, line 763: If you're not using the result from pull.collect, how about just using a pull.drain? pull.collect accumulates an array of all the results, in this case it'll be an array with thousands of `undefined` items. https://github.com/pull-stream/pull-stream/blob/29b4868bb3864c427c3988855c5d65ad5cb2cb1c/sinks/collect.js#L7

Could do like `pull.drain(noop, (err) => {`

staltz (changes requested):

core.js
staltz: Actually, `pull.onEnd((err) => {` seems sweet https://github.com/pull-stream/pull-stream/blob/master/sinks/on-end.js

staltz:
```suggestion
          pull.onEnd((err) => {
```

arj03: Thanks, but this is still around 100s versus 4s

arj03: That might actually be really interesting to try in any case. I'm still trying to understand why this is so much slower, seems like a gotcha in other places as well if pull is this much slower than push.

staltz: Yeah, would be good to learn what the actual perf problem is.

arj03: Yeah. I'm doing this a bit on and off today. I'll report back when I find out more. Thanks for the feedback so far.

arj03: so it turns out to be this one: https://github.com/ssbc/jitdb/issues/189. Maybe we should update jitdb to use a better default of changing it here?

staltz (approved):

arj03:
I pushed up a batch(1000), then the performance is the same as before. We need the same [here](https://github.com/ssbc/ssb-db2/blob/12a0af7d98e87c32b043ab7119298e654180b986/core.js#L986). I'll do that as seperate PR.

PR #405, opened 2022-11-08
Batch 1000 to speed up reindexEncrypted

arj03:
Similar to #403


staltz (approved):
Looks good. Could you also change pull.collect to pull.onEnd? This way it should consume less mem.

arj03:
This should help a lot in tribes2 :)

PR #408, opened 2022-11-29
Recommend toPullStream over toCallback

staltz:
Just a quick tip for other developers in the future, that toCallback can be an OOM footgun.


arj03 (approved):
nice 👍

PR #409, opened 2022-11-29
avoid pull.collect to avoid OOM crashes

staltz:
## Context

Out-of-memory crashes in Manyverse.

## Problem

`pull.collect` considered harmful for RAM usage, because it internally accumulates an array of all stream items.

## Solution

In a few cases, we don't strictly need `pull.collect`, and `pull.drain` / `pull.onEnd` are just fine.


arj03 (approved):
Great, nice find! 👍

PR #410, opened 2022-12-05
Add decrypted api for for box2 decrypted messages

### LINKED ISSUE #406: Live stream needed for reindexEncrypted results
More context at: https://github.com/ssbc/ssb-meta-feeds/pull/110

arj03:
This should fix #406

I went through a bit different solutions (see https://github.com/ssbc/jitdb/pull/236), but ended up with a new pull source api.

I considered doing this so the operators could be reused, but decided against that because the filtering is often pretty simple and this also shows people that doing a filter on the result is not that scary. Where this could also be useful is if you have something like do a rough filter on post but then filter the result only for specific stuff without creating an index in jitdb.

### DIFF
--- a/README.md
+++ b/README.md
@@ -473,14 +473,18 @@ operators:
 See [jitdb operators] and [operators/index.js] for a complete list of supported
 operators.
 
-### decrypted
+### reindexed()
 
-A pull-stream source with newly decrypted values returned as full
-messages. JITDB doesn't have a concept of exactly what values are
-decrypted so a separate api is needed. This api can be combined with
-`where` to receive old values, live values or decrypted values. 
+A pull-stream source of newly decrypted reindexed values returned as
+full messages. Calling `reindexEncrypted` after adding a new box2 key
+will trigger new values in this stream.
 
-Example:
+This api can be combined with `where` to receive messages of a
+particular type no matter if they are existing, newly added or
+decrypted values.
+
+Example of getting existing post messages together with newly
+decrypted ones:
 
 ``` js
     const pull = require('pull-stream')
@@ -493,7 +497,7 @@ Example:
           toPullStream()
         ),
         pull(
-          sbot2.db.decrypted,
+          sbot2.db.reindexed(),
           pull.filter((msg) => {
             return msg.value.content.type === 'post'
           })
--- a/README.md
+++ b/README.md
@@ -602,6 +602,8 @@ at runtime and that changes what messages can be decrypted. Calling
 this function is needed after adding a new key. The function can be
 called multiple times safely.
 
+The decrypted values from this can be consumed using `reindexed`.
+
 ### logStats(cb)
 
 Use [async-append-only-log]'s `stats` method to get information on how many
--- a/README.md
+++ b/README.md
@@ -473,6 +473,40 @@ operators:
 See [jitdb operators] and [operators/index.js] for a complete list of supported
 operators.
 
+### decrypted
+
+A pull-stream source with newly decrypted values returned as full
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -473,6 +473,40 @@ operators:
 See [jitdb operators] and [operators/index.js] for a complete list of supported
 operators.
 
+### decrypted
+
+A pull-stream source with newly decrypted values returned as full
```
staltz, line 478: The implementation does something slightly different, it only emits decrypted values from reindexEncrypted, not all the other conventional decryptions.

I have two ideas:

1. Rename this to reflect that it is only for reindexEncrypted decryptions. I'm also worried that the name `decrypted` is too similar to `operators.js` / `decrypted`
2. What if `reindexEncrypted` *returned* a pull-stream of all the decryptions that occurred in that specific call?

I like option 2, but I'm not sure if it's going to be more developer friendly or less developer friendly.

core.js
```diff
@@ -978,6 +978,22 @@ exports.init = function (sbot, config) {
...
+        if (onValue) onValue(null, val)
+      },
+      source(abort, cb) {
+        if (abort) cb(abort)
+        else if (!onValue) onValue = cb
+      },
+    }
+  }
+
+  const decryptedValues = LazyPull()
```
staltz, line 995: This seems to do the same as `pull-notify`. Could we use that instead? Because it is unit tested and because it evokes a familiar pattern, may helps developers understand what's going on here.

test/query-decrypted-box2.js
```diff
@@ -0,0 +1,109 @@
...
+            sbot2.box2.addGroupInfo(groupId, { key: testkey })
+            sbot2.db.reindexEncrypted((err) => {
+              t.error(err, 'no err')
+            })
+          })
+        )
+      )
+    })
+  })
+})
```
staltz, line 109: The test looks good :)

test/query-decrypted-box2.js
```diff
@@ -0,0 +1,109 @@
...
+    t.true(msgBoxed.value.content.endsWith('.box2'), '.box2')
+
+    const keys2 = ssbKeys.loadOrCreateSync(path.join(dir2, 'secret'))
+    const sbot2 = SecretStack({ appKey: caps.shs })
+      .use(require('../'))
+      .call(null, { keys: keys2, path: dir2 })
+
+    // setup decrypted handler
+    pull(
+      cat([
```
staltz, line 67: Do we need this cat? Seems like we could drop the first stream in the cat and tests would still pass

staltz (changes requested):
Structurally looks good but I have a few suggestions for different approaches.

README.md
arj03: The name change is a good idea. About option 2, I'm afraid that might lead to code where you in multiple places would have to handle there values returned. Instead with the stream approach you can more easily combine there where you handle messages of a specific type anyway I think.

core.js
arj03: right, that was the module I was looking for :)

test/query-decrypted-box2.js
arj03: right, I'll simplify that, I used it in the README so people can find that example there now. Come to think of it, I'll also push an update for reindexEncrypted so that better explains how to handle there values once we settle on how to do that :)

arj03:
Should be ready to go now. Thanks for the review. I'll probably squash this as there is a bit too many commits. Happy with the result :)

staltz (approved):
Looks good! Thanks

PR #412, opened 2023-01-08
Adding ooo messages should not update base index

### LINKED ISSUE #395: `addOOO` should not update Base index
> In general if we do add ooo messages then it should not update this index.

From https://github.com/ssbc/ssb-ebt/pull/72#issuecomment-1274832766

Same should happen with addTransaction because it uses addOOO under the hood, I suppose.

arj03:
This should fix #395. The base index is both used for EBT & validation so doesn't make sense to update for ooo messages.


staltz (approved):
Looks perfect, thanks!

PR #413, opened 2023-01-12
Disable compaction in browser

arj03:
Related to https://github.com/ssbc/async-append-only-log/pull/91. I tried the minimum possible to get things running again. A better fix would of course be to allow compaction to run in the browser, but for now I'd like to get things running again. See https://github.com/arj03/ssb-browser-core/issues/71.

### DIFF
--- a/core.js
+++ b/core.js
@@ -1070,6 +1070,9 @@ exports.init = function (sbot, config) {
 
     if (compacting.value !== !stats.done) compacting.set(!stats.done)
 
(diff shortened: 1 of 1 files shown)


staltz (approved):
It's fine by me if it's fine by you. :)

arj03:
Will probably do some more testing around the new group stuff once we get there, but this is a start. Thanks for the quick review.

PR #417, opened 2023-01-24
deleteFeed to support removing OOO

staltz:
## Context

https://github.com/ssbc/ssb-replication-scheduler/issues/14

## Problem

`deleteFeed` doesn't support deleting messages that were added with OOO, because we have a `state` check, and addOOO has been recently updated #412 to not update the Base `state`.

## Solution

Drop the `state` check in `deleteFeed`.

Add tests.

### DIFF
--- a/core.js
+++ b/core.js
@@ -762,8 +762,6 @@ exports.init = function (sbot, config) {
       stateFeedsReady,
       (ready) => ready === true,
       () => {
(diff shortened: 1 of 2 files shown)


arj03 (approved):
Great, the tests looks good. 

Nice, -2 lines for this "feature" :D

PR #418, opened 2023-01-30
replace clarify-error with new Error cause

staltz:
Note! This makes ssb-db2 incompatible with any version of Node.js smaller than 16.x

I suppose we should mark this as a breaking change.

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -27,8 +27,7 @@ jobs:
 
     strategy:
       matrix:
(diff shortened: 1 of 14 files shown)


arj03:
There is s mention in the readme as well:

Requires Node.js 12 or higher

I guess we didn't need ssb-threads in the tests?

staltz:
Yeah ssb-threads and ssb-friends were used in the benchmarks, but those are `test.skip`ed benchmarks. I think we know what to do later if we want to enable those benchmarks, but I wanted to get rid of the chicken-and-egg with ssb-threads bringing its own old version of ssb-db2.

Will update the README

PR #421, opened 2023-04-10
Actions: update checkout + node-setup verisons

mixmix:
I saw CI complaining:
image

So I took a BLIND STAB at updating the actions ... lets see if they run or if we hit breaking changes to API.

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -17,7 +17,7 @@ jobs:
   licenses:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: REUSE Compliance Check
         uses: fsfe/reuse-action@v1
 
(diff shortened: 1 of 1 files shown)


.github/workflows/node.js.yml
```diff
@@ -60,7 +60,7 @@ jobs:
           body="${body//'%'/'%25'}"
           body="${body//$'\n'/'%0A'}"
           body="${body//$'\r'/'%0D'}"
-          echo ::set-output name=body::$body
+          echo name=body::$body >> $GITHUB_OUTPUT
```
staltz, line 63: This doesn't seem valid. It broke the `benchmark` job

staltz (changes requested):

mixmix:
ok will keep tweaking.
Reference: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

mixmix:
hmmm, error with Node 16 test

image

timing bug

mixmix:
ITS BEAUTIFUL

mixmix:
Referneces: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#multiline-strings

mixmix:
the changes to `sbot.close(t.end)` were because some tests failed with "but the server is already closed". This essentially drops all those errors (which I think it what we want errwhere?

mixmix:
hmmm... passing, but didn't publish benchmark message?

mixmix:
image

staltz (approved):
Yup!

PR #423, opened 2023-05-16
compat/history-stream - add support for { live }

mixmix:
updating scuttle-testbot and noticed that `createHistoryStream({ live: true })` was not supported
Added this along with tests


staltz (approved):
Looks good, thanks!

PR #424, opened 2023-10-02
Add Hook to compat functions

cooldracula:


Powersource:
heey cool :) could we get a test though

Powersource:
or maybe if we do this same pattern in other files it'd maybe not be needed

mixmix:
@Powersource this was Zach and I pairing in this area. Assume he's not coming back to this PR, then decide how you'd like to proceed. If this PR is still needed (I think this mainly exists to make testing some things easier), then can you take it over and pull me if any help needed ?

Powersource:
yeah i think stuff like ssb-recps-guard needs this (if it detects ssb.publish it tries to hook it)

compat/db.js
```diff
@@ -13,21 +14,21 @@ exports.manifest = {
...
       sbot.db.get(idOrObject.id, cb)
     } else {
       sbot.db.get(idOrObject, cb)
     }
-  }
-  sbot.publish = sbot.db.publish
-  sbot.whoami = () => ({ id: sbot.id })
-  sbot.ready = () => true
+  })
+  sbot.publish = Hookable(sbot.db.publish)
```
Powersource, line 27: sbot.db.publish is created by the publish compat plugin. i wonder if this throws if we don't load that first?

Powersource: made a new PR with some changes, didn't want to juggle the branches around https://github.com/ssbc/ssb-db2/pull/427

Powersource:
v8.1.0

PR #425, opened 2023-10-26
Add hasDiskAccess config

Powersource:
In ssb-tribes2-demo we currently have to patch ssb-db2 to get it to run, see https://github.com/ssbc/ssb-tribes2-demo/issues/11 https://github.com/ssbc/ssb-tribes2-demo/blob/3d4425fec46894b4c0db601c312b65e9d58fe9ae/patches/ssb-db2%2B7.1.1.patch This config option should allow us to remove that patch.

I didn't add a test here but I'll make a PR in ssb-tribes2-demo where i test it

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,8 +21,8 @@ Over time, this database received more features than ssb-db, and now supports:
 - Deletion and compaction
 - Customizable feed formats and encryption formats
   - You are not tied to classic SSB messages and the classic mode of encryption,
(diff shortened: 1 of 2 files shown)


staltz (approved):
Yeah this seems okay.

Powersource:
v7.2.0

Powersource:
woops i think setting this to `false` doesn't force it to be false. but i think the only real usecase i know of is electron, and there you'd want to set it to `true` so no biggie

PR #428, opened 2024-04-11
bulk create adds messages to db in wrong order

mixmix:
Noticed that when adding lots of messages to the database quickly, they go in right,
but when you stream the out of the db, they come out with the wrong sequence order

This is a problem for legacy replication (`createHistoryStream`)


test/create.js
```diff
@@ -85,6 +94,46 @@ test('create() classic "too big"', (t) => {
...
+  const N = 4
+  const many = new Array(N).fill('').map(() => {
+    return { type: 'counter' }
+  })
+
+  const expectedSequenceOrder = new Array(N).fill(0).map((_, i) => i + 1)
+
+  await Promise.all(
+    many.map((content) => {
+      return p(db.create)({ feedFormat: 'classic', content })
```
staltz, line 107: Promise.all() is not the right way of doing this, you need to do them sequentially. Otherwise it's very likely indeed that they'll end up in the wrong order, because this has to do with jitdb indexing the log.

ssb-db2 is optimized for fast reads not for fast writes, so be patient if it's not the best at bulk creating msgs.

mixmix: I wrote a test where I did it sequentially with a pull-stream, and it still got them wrong I think...

the fast writes thing is gonna be a challenge with Ahau importing some records. Working on that though

